PR #30, opened 2014-11-06
Added LinkedIn Provider

otherguy:
I also changed some things in the AbstractProvider class to make it easier to configure new providers instead of completely overriding some methods.

For example, LinkedIn required scopes to be separated by a space and the request URL query parameters need to be encoded according to RFC3986 (spaces converted to `%20` instead of `+`). Other platforms will have different requirements (scopes separated by a comma or by a space but encoded as per RFC1738 etc.).

The `$scope_separator` and `$encoding_type` can now be configured on a per-provider basis.  This change makes it even more trivial to add new providers.

### DIFF
--- a/src/Two/LinkedinProvider.php
+++ b/src/Two/LinkedinProvider.php
@@ -83,7 +83,7 @@ class LinkedinProvider extends AbstractProvider implements ProviderInterface {
 	{
 		return (new User)->setRaw($user)->map([
 			'id' => $user['id'], 'nickname' => null, 'name' => $user['formattedName'],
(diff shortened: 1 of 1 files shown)


AndyWendt:
@darkwinternight  just override them in your implementing class or use a setter or use the constructor.

darkwinternight:
@AndyWendt Override what? `$scopeSeparator`? Or the whole method?

The point of my pull request was to add the `$scopeSeparator` field in the first place. @taylorotwell accepted my pull request here: https://github.com/laravel/socialite/pull/37

PR #38, opened 2015-01-10
Fix import and docblock

joecohens:


src/SocialiteManager.php
```diff
@@ -112,11 +113,11 @@ protected function formatConfig(array $config)
 	/**
 	 * Get the default driver name.
 	 *
-	 * @return string
+	 * @throws InvalidArgumentException
```
GrahamCampbell: we still need the leading slash in the docblock, and we need a return string here still since it still defines the return type

joecohens: Thanks :)

src/SocialiteManager.php
```diff
@@ -112,11 +113,11 @@ protected function formatConfig(array $config)
 	/**
 	 * Get the default driver name.
 	 *
-	 * @return string
+	 * @throws \InvalidArgumentException
 	 */
```
GrahamCampbell: we still need a return annotation

joecohens: What we should return `string` or `void`?

GrahamCampbell: string

PR #39, opened 2015-01-24
Added BitBucket Provider

joecohens:


GrahamCampbell:
:+1:

DmitryRomanenko:
I think this provider should be in different repository because it's not popular with end users.

joecohens:
I don't agree most services handling GitHub handle BitBucket also. Even Laravel Forge had lots of request one Taylor launched to support BitBucket.

JulienTant:
I'm not sure it's a good idea to manage a lot of providers by default, because it's such a pain then to maintain every single provider changes.

joecohens:
I guess you're right but the OAuth1 providers in this package are managed by https://github.com/thephpleague/oauth2-client https://github.com/laravel/socialite/blob/2.0/composer.json#L18

GrahamCampbell:
I still think this provider is good to have in the core.

antonioribeiro:
Bitbucket is the way to go if you need private repositories and cannot afford a Github private account. +1

AndyWendt:
We added a bitbucket provider (as well as a ton of others) here: https://github.com/SocialiteProviders/BitBucket

taylorotwell:
This could be sent to 2.0 branch.

joecohens:
See https://github.com/laravel/socialite/pull/54

PR #40, opened 2015-01-24
Add email to github user data

DmitryRomanenko:
Github didn't return email field before.  
Even if `$scopes` has `user:email` entry.

### DIFF
--- a/src/Two/GithubProvider.php
+++ b/src/Two/GithubProvider.php
@@ -32,13 +32,26 @@ class GithubProvider extends AbstractProvider implements ProviderInterface {
 	 */
 	protected function getUserByToken($token)
 	{
-		$response = $this->getHttpClient()->get('https://api.github.com/user?access_token='.$token, [
+		$options = [
 			'headers' => [
 				'Accept' => 'application/vnd.github.v3+json',
 			],
-		]);
+		];
 
-		return json_decode($response->getBody(), true);
+		$response = $this->getHttpClient()->get('https://api.github.com/user?access_token='.$token, $options);
+		$user = json_decode($response->getBody(), true);
+
+		if (in_array('user:email', $this->scopes)) {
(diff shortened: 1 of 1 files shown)


src/Two/GithubProvider.php
```diff
@@ -32,13 +32,26 @@ protected function getTokenUrl()
...
 			'headers' => [
 				'Accept' => 'application/vnd.github.v3+json',
 			],
-		]);
+		];
+
+		$response = $this->getHttpClient()->get('https://api.github.com/user?access_token='.$token, $options);
+		$user = json_decode($response->getBody(), true);
+
+		if (in_array('user:email', $this->scopes)) {
```
GrahamCampbell: Braces need to go on their own line.

src/Two/GithubProvider.php
```diff
@@ -32,13 +32,26 @@ protected function getTokenUrl()
...
+			$response = $this->getHttpClient()->get('https://api.github.com/user/emails?access_token='.$token, $options);
+			$emails = json_decode($response->getBody(), true);
+			foreach ($emails as $email) {
```
GrahamCampbell: Same issue.

src/Two/GithubProvider.php
```diff
@@ -32,13 +32,26 @@ protected function getTokenUrl()
...
+				if ($email['primary'] && $email['verified']) {
```
GrahamCampbell: And here.

src/Two/GithubProvider.php
```diff
@@ -32,13 +32,26 @@ protected function getTokenUrl()
...
+					$user['email'] = $email['email'];
```
GrahamCampbell: Shouldn't we break out of the loop once we've added an email address? Minor optimization. :)

GrahamCampbell:
Replaced by #41. Please close this.

joecohens:
:+1:

PR #41, opened 2015-01-26
Add Email To GitHub User Data

GrahamCampbell:
Replaces #40 with some fixes and an early exit optimization.

### DIFF
--- a/src/Two/GithubProvider.php
+++ b/src/Two/GithubProvider.php
@@ -32,13 +32,26 @@ class GithubProvider extends AbstractProvider implements ProviderInterface {
 	 */
 	protected function getUserByToken($token)
 	{
-		$response = $this->getHttpClient()->get('https://api.github.com/user?access_token='.$token, [
+		$options = [
 			'headers' => [
 				'Accept' => 'application/vnd.github.v3+json',
 			],
-		]);
+		];
 
-		return json_decode($response->getBody(), true);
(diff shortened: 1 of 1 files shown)


joecohens:
:+1:

metekabak:
Will someone merge this solution? We still can't get github private email(s).

IsraelOrtuno:
Have tried this code and works perfect, is it going to be merged?

amirhabibzadeh:
It works great , thanks . is it going to be merged to the main repo?

taylorotwell:
Probably should be sent to 2.0 branch.

DmitryRomanenko:
Why not to master?
Github provider doesn't work without this patch (email is always null).

taylorotwell:
Because master isn't the current release? 2.0 is.

taylorotwell:
Can someone re-target this?

DmitryRomanenko:
@taylorotwell I will create a new one from scratch (because code style has been changed) and I'll target it to 2.0.

GrahamCampbell:
You should retain the original commits.

DmitryRomanenko:
> You should retain the original commits.
I won't use that commits or code from these commits.

DmitryRomanenko:
> You should retain the original commits.

The problem is I don't know how to retain original commits because I want to use a new base commit (it's better to start with current 2.0 branch).

PR #43, opened 2015-01-28
Added methods to factory interface for preventing 'Method *** not found'...

DmitryRomanenko:
Added methods to factory interface for preventing 'Method *** not found' in IDEs.

It's useful for ServiceProviders that add custom drivers using `extend`, etc...


taylorotwell:
Including __call an interface doesn't make sense.

PR #47, opened 2015-02-04
Remove duplicate src/Two/ProviderInterface

morrislaptop:
These providers already implement AbstractProvider which implements
Laravel\Socialite\Contracts\Provider as ProviderContract


morrislaptop:
?

PR #50, opened 2015-02-13
Add Dropbox provider

jasonvarga:
Needed to override `getAccessToken` because apparently Dropbox doesn't like when you send the request as JSON.
Also had to override `getTokenFields` because they require `grant_type` in there.

They don't have avatars so I used Gravatar. Not sure if that's a good or bad thing.

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -6,6 +6,7 @@ use Laravel\Socialite\Two\GithubProvider;
 use Laravel\Socialite\Two\GoogleProvider;
 use Laravel\Socialite\One\TwitterProvider;
 use Laravel\Socialite\Two\FacebookProvider;
+use Laravel\Socialite\Two\DropboxProvider;
 use League\OAuth1\Client\Server\Twitter as TwitterServer;
 use Laravel\Socialite\One\AbstractProvider as AbstractOneProvider;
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
:-1: This is overkill and doesn't belong in the core.

peter-mw:
Dropbox is nice to have, @GrahamCampbell why do you think its overkill?

the more providers the more the merrier

If we have the popular providers in the core this means easier integration of services

AndyWendt:
This is great to have a Dropbox provider  :smile:  though I believe that @taylorotwell said that he was going to stick with the 4 main providers only.  So I think it will be :-1:

GrahamCampbell:
Check out https://github.com/DraperStudio-Socialite-Provider.

AndyWendt:
I did though it doesn't look like he extended socialite.  It looks like you have to remove the `SocialiteServiceProvider` in order to use his which is stupid.  I will play around with it.

DraperStudio:
I pushed an update for all Providers so that they use the `extend`-method instead of replacing the Socialite Manager/ServiceProvider.

AndyWendt:
Awesome!

jasonvarga:
I had no idea about that package.

taylorotwell:
Should be able to use `extend` method in Socialite manager to add providers for your app.

PR #54, opened 2015-02-23
Added BitBucket provider

joecohens:


src/One/BitbucketProvider.php
```diff
@@ -0,0 +1,7 @@
+<?php namespace Laravel\Socialite\One;
+
+class BitbucketProvider extends AbstractProvider
+{
+
```
GrahamCampbell: one too many blank lines here :)

joecohens: I know :( just following this package conventions https://github.com/laravel/socialite/blob/2.0/src/One/TwitterProvider.php

GrahamCampbell: Looks like a mistake when Taylor ran the automatic psr-2 conversion.

PR #56, opened 2015-02-23
Fix for undefined index error when email is hidden

dexbarrett:
Same fix as the FB Provider. In Github (at least with new accounts) the email is set to be hidden by default in the user profile and this causes an error since that field won't be available for Socialite's **mapUserToObject** method.

Undefined index error before fix:

socialite-github-noemail-error

Fallback null value for missing email after fix:

socialite-github-fix

### DIFF
--- a/src/Two/GithubProvider.php
+++ b/src/Two/GithubProvider.php
@@ -49,7 +49,7 @@ class GithubProvider extends AbstractProvider implements ProviderInterface
     {
         return (new User)->setRaw($user)->map([
             'id' => $user['id'], 'nickname' => $user['login'], 'name' => array_get($user, 'name'),
-            'email' => $user['email'], 'avatar' => $user['avatar_url'],
+            'email' => isset($user['email']) ? $user['email'] : null, 'avatar' => $user['avatar_url'],
         ]);
     }
 }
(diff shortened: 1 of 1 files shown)


joecohens:
You can use the `array_get` like name does.

bexdarrett:
@joecohens Right! Fixed :D

dexbarrett:
@joecohens Right! Fixed :D

courtaks:
aha! finally found the answer..
please add it to the final version...

dincio:
im using facebook provider but how can i get the location and birthday of the user? i can only get right now are emails and name. thanks

courtaks:
socialite is only for signing in users with these providers. if you need any thing else please go to facebook's api & read the documentation: 
https://developers.facebook.com/docs/apis-and-sdks

dexbarrett:
@dincio The info you need from the user can be specified  with the `scopes` method http://laravel.com/docs/5.0/authentication#social-authentication However, those permissions are not granted by default to your Facebook app; when your app requests those kind of permissions it must be reviewed by Facebook first:

>Review

>If your app asks for more than than public_profile, email and user_friends it will require review by Facebook before your app can be used by people other than the app's developers.

>The time to review your app is usually about 7 business days. Some extra-sensitive permissions, as noted below, can take up to 14 business days.

https://developers.facebook.com/docs/facebook-login/permissions/v2.2

courtaks:
can you add linked-in login as well...
https://developer.linkedin.com/docs/signin-with-linkedin

PR #60, opened 2015-03-08
getting user data using facebook login

dincio:
just wanna ask how to get the user's facebook data like user_birthday, user_location?
the data i can only get now is the name and email of user's facebook data when they login.


taylorotwell:
Make an API request using the token you get.

PR #61, opened 2015-03-09
Duplicate entry

dincio:
I implement socialite together with manual registration where register using facebook is a user's option. but i noticed 
that if the user registered first with the manual registration and logout then came back and register using facebook it 
produce an error  "SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry"

here is my current UserRepository

namespace App\Repositories;
use App\User;
class UserRepository{
    public function findByUsernameOrCreate($userData)
    {
            return User::firstOrCreate([
                'first_name' => $userData->first_name,
                'last_name' => $userData->last_name,
                'avatar' => $userData->avatar,
                'gender' => $userData->gender,
                'email' => $userData->email
            ]);
    }
}

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-master": "3.0-dev"
         }
     },
     "minimum-stability": "dev"


GrahamCampbell:
?

matinwd:
Still have this problem
Any solution?

PR #62, opened 2015-03-09
Added redirect($rerequest) option to force permission screen

nlogachev:
This PR adds the ability to force the permissions screen again in Facebook and Google. Socialite as is doesn't allow for this behaviour. If a user allows the app access but drops, for example, the `email` permission in the process, Socialite becomes a bit useless.

It's a drop in that doesn't change the established behaviour of Socialite installations.

You would use it like so: `return $socialite->with($provider)->redirect(true);`

This still works as expected: `return $socialite->with($provider)->redirect();`

Additional unit tests included specifically for this feature.

### DIFF
--- a/src/Contracts/Provider.php
+++ b/src/Contracts/Provider.php
@@ -6,9 +6,11 @@ interface Provider
     /**
      * Redirect the user to the authentication page for the provider.
      *
+     * @param bool $rerequest Whether or not to rerequest permissions.
+     *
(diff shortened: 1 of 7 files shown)


taylorotwell:
One issue is this breaks all custom Socialite providers people may be using since it adds a new parameter to the method. So, if this is done, it would have to on the next major release of Socialite.

nlogachev:
@taylorotwell :+1:

PR #64, opened 2015-03-26
add setCode method

liaol:
Add setCode method
You may need set code inputed from moblie client instead of redirecting when write api.


taylorotwell:
Formatting is pretty bad here.

PR #66, opened 2015-03-31
Add Email To GitHub User Data

DmitryRomanenko:
Github didn't return email field before.

Even if $scopes has user:email entry.

Replaces #40, #41

### DIFF
--- a/src/Two/GithubProvider.php
+++ b/src/Two/GithubProvider.php
@@ -1,7 +1,5 @@
 <?php namespace Laravel\Socialite\Two;
 
-use Symfony\Component\HttpFoundation\RedirectResponse;
-
 class GithubProvider extends AbstractProvider implements ProviderInterface
 {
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
In my tests it does: http://d.pr/i/FaCk/1YabVHQD

Do I need to do something special to make it not return the e-mail address?

DmitryRomanenko:
Github:
g1795047@trbvm.com
111222zzz

I've just created a new login and it doesn't have any email field.
Also it doesn't have a name.

taylorotwell:
E-Mail still returns as `null` even after applying your changes.

GrahamCampbell:
@taylorotwell My original pull worked fine?

GrahamCampbell:
We were using it in prod on styleci for a while.

dexbarrett:
@taylorotwell @GrahamCampbell @DmitryRomanenko I don't know if this is the case, but I noticed before that new GitHub accounts have the email set to hidden by default. I think that's why yours might be returning null. It happened to me and I just set it to be visible on the profile settings https://github.com/settings/profile

dimsav:
> Some GitHub accounts have the email set to hidden.

I confirm that when the email is set to hidden, the API doesn't provide the email of the user.

GrahamCampbell:
>  the API doesn't provide the email of the user.

It does if you set the correct scopes.

sitesense:
I confirm the error. With existing code I get this on the callback:
~~~
SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'email' cannot be null (SQL: insert into `users` (`name`, `email`, `password`, `updated_at`, `created_at`) values (Mark, , yURj/23m3ZUxmsl/9JYIrunz4GADGvfapXxGxXTtPEVumbLR2v2sq, 2015-05-07 22:55:03, 2015-05-07 22:55:03))
~~~
DmitryRomanenko's patch works for me.

PR #67, opened 2015-04-13
Allow for implicit grants and supplementary redirect_uris.

atrauzzi:


atrauzzi:
Talked about here: https://github.com/laravel/framework/issues/8398

src/Two/AbstractProvider.php
```diff
@@ -177,6 +177,16 @@ protected function hasInvalidState()
     {
         $session = $this->request->getSession();
 
+        if($this->request->has('code')) {
+
+            if($redirectUri = $this->request->get('redirect_uri')) {
```
GrahamCampbell: cs

src/Two/AbstractProvider.php
```diff
@@ -177,6 +177,16 @@ protected function hasInvalidState()
     {
         $session = $this->request->getSession();
 
+        if($this->request->has('code')) {
```
GrahamCampbell: cs

atrauzzi:
Sorry, CS?

taylorotwell:
It feels weird to put this in getInvalidState(), since it has nothing to do with determining if the "state" is invalid. Could we do this differently?

atrauzzi:
We could possibly do it by using `getInvalidState` in combination with other things that "okay" the current request.  The problem being, the current code insists that getInvalidState be...`false`?  So, we'd either be doing an `||`, or making another function like "isCurrentRequestOk".

taylorotwell:
I think you can use the new stateless() method on the provider.

PR #70, opened 2015-05-11
Add eve online SSO service

OrthoLoess:
Unsure if this is wanted in the main repo, so just close if not.

EVE Online has an OAuth 2 SSO implementation, used by a lot of player-built sites and tools.


taylorotwell:
You can use the `extend` method on your manager to add your own drivers.

PR #71, opened 2015-05-20
Added getOAuthVersion() method to AbstractProvider classes.

shehi:
I added ```getOAuthVersion()``` method to AbstractProvider classes and related Contract. This method returns ```1``` for v1 implementations (namely Twitter) and ```2``` for their v2 counterparts. Additionally, I fixed indentation from spaces to tabs, as Laravel project requires.

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -1,5 +1,6 @@
 <?php namespace Laravel\Socialite\Two;
 
+use GuzzleHttp\Client;
 use Illuminate\Http\Request;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Laravel\Socialite\Contracts\Provider as ProviderContract;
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Please revert all changes to the cs.

shehi:
@GrahamCampbell : What is CS?

shehi:
CS = coding standards. Done.

GrahamCampbell:
This is a breaking change to and interface so should go to the master IMO.

src/Two/AbstractProvider.php
```diff
@@ -265,11 +266,11 @@ public function scopes(array $scopes)
     /**
      * Get a fresh instance of the Guzzle HTTP client.
      *
-     * @return \GuzzleHttp\Client
+     * @return Client
```
GrahamCampbell: Keep the FQCN in the phpdoc always please.

PR #74, opened 2015-06-09
Added support for Facebook popup dialog

mirzabusatlic:
When running the callback in a popup, Facebook recommends using ```&display=popup``` in order for the dialog to render properly.


mirzabusatlic:
Thanks Taylor. Can you draft a new release for this?

PR #77, opened 2015-06-18
Add LinkedIn Provider

barryvdh:
This adds a LinkedIn Provider with basis profile + email fields.
Besides the name/email/avatar, some additional raw fields are requested.

This currently doesn't work on the latest Guzzle 6 release, until a release with https://github.com/guzzle/psr7/commit/962735c754f2a562a619b456a3aa4e614f0e5360 is tagged.

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -69,6 +69,20 @@ class SocialiteManager extends Manager implements Contracts\Factory
     }
 
     /**
+     * Create an instance of the specified driver.
+     *
+     * @return \Laravel\Socialite\Two\AbstractProvider
+     */
+    protected function createLinkedinDriver()
+    {
+        $config = $this->app['config']['services.linkedin'];
+
+        return $this->buildProvider(
+          'Laravel\Socialite\Two\LinkedInProvider', $config
+        );
+    }
+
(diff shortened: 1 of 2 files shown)


src/Two/LinkedinProvider.php
```diff
@@ -0,0 +1,75 @@
+<?php namespace Laravel\Socialite\Two;
+
+use GuzzleHttp\ClientInterface;
+
+class LinkedinProvider extends AbstractProvider implements ProviderInterface
```
dwightwatson: Any reason this wouldn't be `LinkedInProvider`?

barryvdh: Could also be.

taylorotwell: Yeah let's go with a capital I.

barryvdh: But driver name still just 'linkedin' right?

taylorotwell: Right.

barryvdh: Fixed and squashed.

taylorotwell:
Thanks :)

PR #78, opened 2015-06-22
Add original picture to LinkedIn

barryvdh:
Make the fields a bit readable + include the original picture in the profile response.
The regular (small) picture is still used for the default avatar.

### DIFF
--- a/src/Two/LinkedInProvider.php
+++ b/src/Two/LinkedInProvider.php
@@ -1,7 +1,5 @@
 <?php namespace Laravel\Socialite\Two;
 
-use GuzzleHttp\ClientInterface;
-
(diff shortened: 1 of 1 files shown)


src/Two/LinkedInProvider.php
```diff
@@ -33,9 +42,10 @@ protected function getTokenUrl()
      */
     protected function getUserByToken($token)
     {
-        $fields = 'id,email-address,first-name,last-name,formatted-name,headline,picture-url,public-profile-url,location';
+        $fields = implode(',', $this->fields);
```
taylorotwell: Need a blank line in between these two lines.

barryvdh: Added.

PR #79, opened 2015-06-22
Add original/large avatars

barryvdh:
This adds an attribute `avatar_original` with the larger images for LinkedIn, Twitter and Facebook.
The normale avatar pictures are pretty small, so if you want to use them for anything else then a tiny thumb, this makes that possible.

LinkedIn: Supplies original image when request
Twitter: Suggests so strip `_normal`: https://dev.twitter.com/overview/general/user-profile-images-and-banners
Facebook: Has a width/height parameter, just supply a large constraint (1920) in this case)

Perhaps other providers have the same possibilities, but not sure. If more support this, we could add a method to the AbstractUser perhaps?

### DIFF
--- a/src/One/TwitterProvider.php
+++ b/src/One/TwitterProvider.php
@@ -19,6 +19,7 @@ class TwitterProvider extends AbstractProvider
         return $instance->map([
             'id' => $user->uid, 'nickname' => $user->nickname,
             'name' => $user->name, 'email' => $user->email, 'avatar' => $user->imageUrl,
+            'avatar_original' => str_replace('_normal', '', $user->imageUrl),
         ]);
     }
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
Have you tested these on these method on the various providers?

barryvdh:
I tested these 3 with my own account. Didn't test any other providers yet.

barryvdh:
Github already gives a large image I think (at least larger than tiny)
Bitbucket can replace `/32/` with `/1920/` to change the size.

Should I add a `getOriginalAvatar()` method to the AbstractUser?

nkuldip:
How to increase Google user profile size. 
https://lh5.googleusercontent.com/-34GzH9xpFbA/AAAAAAAAAAI/AAAAAAAAAQU/WJBzTU1Lw58/photo.jpg?sz=50

barryvdh:
I'm guessing, change the `?sz=50' to something larger ;)
https://lh5.googleusercontent.com/-34GzH9xpFbA/AAAAAAAAAAI/AAAAAAAAAQU/WJBzTU1Lw58/photo.jpg?sz=800

nkuldip:
Yes but how?

PR #82, opened 2015-06-29
Loosen the auth requirements

PaddoSwam:
This change loosens the requirements by asking just email/basic info instead of full Google+ bulk data which isn't used the Socialite driver.

### DIFF
--- a/src/Two/GoogleProvider.php
+++ b/src/Two/GoogleProvider.php
@@ -75,7 +75,7 @@ class GoogleProvider extends AbstractProvider implements ProviderInterface
      */
     protected function getUserByToken($token)
     {
-        $response = $this->getHttpClient()->get('https://www.googleapis.com/plus/v1/people/me?', [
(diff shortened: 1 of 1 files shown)


taylorotwell:
Is this a breaking change?

PaddoSwam:
I might have submitted it too early as I have not tested if this yield the same information, but I think this is all that's needed! Let me make a test for the mapUserToObject function.

p.s. for what it's worth 'it works for me' ;)

PaddoSwam:
Yes and no.

Unchanged values:
**id**
**name**

Changed values:
**nickname** (not available mapped name)
**avatar** (but added the difference ?sz=50)
**email** (in the original code I didn't have an email address, with the lesser requirements I do receive an email address)

For testing the different url's and their responses also see [Google Oauth Playground](https://developers.google.com/oauthplayground/)

Personally I prefer the '*just email + name*' vs. '*google+ profile*' requirement, but I see how it could potentially break existing sites and applications.

PR #86, opened 2015-06-30
Add refreshToken fetch and action

Korko:
Add two fields in the user return: refreshToken and tokenExpiresIn so that if requested, these tokens can be used later on.
Add method in OAuth2 providers to refresh an accessToken from a refreshToken.


taylorotwell:
Holding off on this for now.

Korko:
Ok, anyway, it was missing the new accessToken in the return of the refresh method ;)

PR #87, opened 2015-07-03
throw a Facebook exception

hex7c0:
Guzzle throw a unintelligible Exception, if status code is different from 200
`Client error: 400`


taylorotwell:
You can log the full error in your application.

hex7c0:
but not the body of facebook request

PR #90, opened 2015-07-09
Adds 'fields' parameter support

joebenson:
Graph API v2.4 requires a fields parameter to return anything beyond id, name, and avatar. This fix sets the default fields to match the default scope request and common user mapping (first_name, last_name, email). Also allows custom fields to be requested instead via the fields() method.


ChrisReid:
I had just done the EXACT same thing right down to the $fields property. Spent the last two hours pulling my hair out and just fixed it. Was just about to open a pull request but you beat me to it by an hour :+1: 
Anyway, your fix works for me.

PR #92, opened 2015-07-22
Facebook Login is not returning some fields

chako91:
When login with Facebook them is no returning fields: gender and verified.


taylorotwell:
Does this require additional scopes?

chako91:
Im using it without scopes. Ones of the lassts fiexies broke my code, so i fix it adding those fields (gender and verified), i dont know if there are more missing.

PR #93, opened 2015-07-30
Pull the state from the session

GrahamCampbell:
By doing this we're protecting against people trying to guess the state.

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -202,7 +202,7 @@ abstract class AbstractProvider implements ProviderContract
 
         $state = $this->request->getSession()->pull('state');
 
-        return strlen($state) !== 40 || $this->request->input('state') !== $state;
+        return ! (strlen($state) > 0 && $this->request->input('state') === $state);
     }
 
     /**
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -202,7 +202,7 @@ abstract class AbstractProvider implements ProviderContract
 
         $session = $this->request->getSession();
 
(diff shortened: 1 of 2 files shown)


src/Two/AbstractProvider.php
```diff
@@ -200,9 +200,9 @@ protected function hasInvalidState()
             return false;
         }
 
-        $session = $this->request->getSession();
+        $state = $this->request->getSession()->pull('state');
 
-        return ! ($this->request->input('state') === $session->get('state'));
+        return ! (strlen($state) > 0 && $this->request->input('state') === $state);
```
mfdj: Sorry to swoop in, I found this PR via [FriendsOfPhp](https://github.com/FriendsOfPHP/security-advisories/pull/78/files) and was just glancing around and wonder if the above comparison is susceptible to a [timing attack](https://en.wikipedia.org/wiki/Timing_attack)? If an attacker can control `$this->request->input('state')` then strict string equality may be leaking information.

taylorotwell: The "pull" method on the session removed state after its retrieved. So you
can only attempt to guess any given state exactly one time. So, timing
attack wouldn't be possible here since you can only try once.

GrahamCampbell: It's not susceptible to a timing attacl.

GrahamCampbell: That's exactly what this PR fixes actually. "pull" from the session means we only read the value once, then it's gone, so someone can't try to guess it again.

GrahamCampbell: Didn't see your comment there Taylor. :)

PR #104, opened 2015-09-22
Allow for custom config

vmichnowicz:
The [JIRA Socialite Provider](https://github.com/SocialiteProviders/Jira) has a hardcoded JIRA base URL of "http://example.jira.com". Since JIRA is typically hosted on various company servers there is no one official URL that can be placed in here. There should be some clean way of setting this URL via config. But since the `formatConfig` method strips out all custom configs I needed to make some updates so I can pass the JIRA URL to the provider.

I submitted [a pull request](https://github.com/SocialiteProviders/Jira/pull/1) to the JIRA Socialite Provider repo to grab the custom config and populate that URL.

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -129,15 +129,14 @@ class SocialiteManager extends Manager implements Contracts\Factory
      */
     public function formatConfig(array $config)
     {
-        $return = array();
-        $custom = array('client_id' => 'identifier', 'client_secret' => 'secret', 'redirect' => 'callback_uri');
+        $return = [];
+        $custom = ['client_id' => 'identifier', 'client_secret' => 'secret', 'redirect' => 'callback_uri'];
 
         foreach ($config as $key => $value) {
             // If there is a custom mapping for this config
             if (isset($custom[$key])) {
                 $return[ $custom[$key] ] = $value;
-            }
-            else {
+            } else {
                 $return[$key] = $value;
             }
         }
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -129,11 +129,20 @@ class SocialiteManager extends Manager implements Contracts\Factory
      */
     public function formatConfig(array $config)
     {
-        return [
-            'identifier' => $config['client_id'],
-            'secret' => $config['client_secret'],
-            'callback_uri' => $config['redirect'],
-        ];
+        $return = array();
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Please fix the cs issues.

taylorotwell:
Couldn't this just be an `array_merge` or something?

vmichnowicz:
Are you saying something like this?

```
return array_merge([
	'identifier' => $config['client_id'],
	'secret' => $config['client_secret'],
	'callback_uri' => $config['redirect'],
	], $config);
```

I agree that is much cleaner. It does result in the same config value getting included twice. I got no problem with that if it sounds good to you.

```
...
'identifier' => 'my_identifier',
'client_id' => 'my_identifier',
...
```

taylorotwell:
What do you mean by it results in the same config value getting included twice?

vmichnowicz:
With the `array_merge` some of the config values would unnecessary end up getting included twice. This code may explain what I am trying to say:

```php
<?php

$config = [
    'client_id' => 'my-identifier-id',
    'client_secret' => 'my-secret',
    'redirect' => 'my-callback-uri',
    'custom' => 'my-custom-config',
];

function formatConfig(array $config) {
    return [
        'identifier' => $config['client_id'],
        'secret' => $config['client_secret'],
        'callback_uri' => $config['redirect'],
    ];
}

function formatConfigTwo(array $config) {
    return array_merge([
        'identifier' => $config['client_id'],
        'secret' => $config['client_secret'],
        'callback_uri' => $config['redirect'],
    ], $config);
}

// The current formatConfig method strips out the custom config I need to send to JIRA Socialite
var_dump(formatConfig($config));

// array (size=3)
//   'identifier' => string 'my-identifier-id'
//   'secret' => string 'my-secret'
//   'callback_uri' => string 'my-callback-uri'

// An array_merge results in some of the config being duplicated
var_dump(formatConfigTwo($config));

// array (size=7)
//   'identifier' => string 'my-identifier-id'
//   'secret' => string 'my-secret'
//   'callback_uri' => string 'my-callback-uri'
//   'client_id' => string 'my-identifier-id' <<< Duplicate
//   'client_secret' => string 'my-secret'    <<< Duplicate
//   'redirect' => string 'my-callback-uri'   <<< Duplicate
//   'custom' => string 'my-custom-config'
```

taylorotwell:
Fixed.

PR #107, opened 2015-10-20
Fix Facebook name lexical order problem.

taichunmin:
For most Chinese, Japanese and Korean people, lexical order of name is "{last}{first}". Facebook user can customize the order in Facebook settings. Instead of use "first_name" and "last_name", we can just change to use "name" fields. This changes can fix lexical order of name.


taylorotwell:
Done on 2.0 branch.

taichunmin:
Good job!  074e3cb685ab3e9a9c31752e48c0235110938b98

PR #110, opened 2015-11-03
Replaced G+ scopes by Google Sign

Max13:
When using Google Sign-in, using G+ is unnecessary and can be blocked with Google Apps.
Google+ can be disabled on Google Apps domains, or when enabled, it forces the user to subscribe to G+.


Max13:
@taylorotwell Hi again. I may have missed something, is it normal not have this pull request included when composer install it? Is it related to tags not being updated?

PR #111, opened 2015-11-06
Make AbstractProvider::scopes() more intuitive

KerryJones:
Google and Facebook require certain scopes to login correctly so they
don't throw an error, which are included by default. However, when a
user adds their own scopes, it overrides these scopes that they don't
know about. This makes it merge them instead.

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -275,9 +275,19 @@ abstract class AbstractProvider implements ProviderContract
      */
     public function scopes(array $scopes)
     {
-        $this->scopes = $scopes;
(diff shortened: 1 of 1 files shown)


src/Two/AbstractProvider.php
```diff
@@ -275,10 +275,20 @@ protected function getCode()
      */
     public function scopes(array $scopes)
     {
-        $this->scopes = $scopes;
+        $this->scopes = array_unique(array_merge($this->scopes, $scopes));
 
         return $this;
     }
+
```
GrahamCampbell: cs

GrahamCampbell: See StyleCI's report.

PR #113, opened 2015-11-09
Added OpenID scope

Max13:
Without the `openid` scope, the name of the user isn't returned when G+ is disabled (for Google Apps) or the user isn't registered


tominon:
:thumbsup:

PR #116, opened 2015-11-24
Switch from pull() to get()

slipszenko:
`$this->request->getSession()` seems to be returning `Symfony\Component\HttpFoundation\Session` for us (not `\Illuminate\Session\Store`), which provides no `pull()` method.

Changing this has stopped us from getting the error: "Call to undefined method Symfony\Component\HttpFoundation\Session\Session::pull()", from `vendor/laravel/socialite/src/Two/AbstractProvider.php:212`.

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -209,7 +209,7 @@ abstract class AbstractProvider implements ProviderContract
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Are you using Laravel?

slipszenko:
We have it included in composer for various different parts of the application (our models use Eloquent), but the core of the application isn't a Laravel application.

PR #117, opened 2015-11-24
Added Bitbucket as an OAuth2 provider

mpociot:
The current implementation only uses the OAuth 1 provider which is lacking important information like the UUID. 

Also the OAuth 2 token allows you to clone bitbucket repositories, while this is not possible with the OAuth 1 token.

### DIFF
--- a/src/One/BitbucketProvider.php
+++ b/src/One/BitbucketProvider.php
@@ -1,8 +1,0 @@
-<?php
-
-namespace Laravel\Socialite\One;
-
-class BitbucketProvider extends AbstractProvider
-{
-    //
-}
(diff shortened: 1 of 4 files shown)


taylorotwell:
Wouldn't this be a breaking change. Would go on the 3.0 / master branch.

mpociot:
Yep - Should I reopen this on the master branch? 

I wasn't sure about which branch to use for the PR.

taylorotwell:
Yes. Thanks.

mpociot:
Damn...nevermind - just saw that there was a PR for this 17 days ago which is already merged into master.

PR #121, opened 2015-11-30
Allow named routes to be used for callbacks

CurosMJ:
The callbacks were restricted to fully-qualified URLs before this. This means for an app to generate proper redirect URLs, it would be dependent upon using some env variable to determine the correct base URL.

This PR introduces `UriResolver` which allows mapping of named routes and paths for redirect URLs.


taylorotwell:
Don't want to add router dependency. Sorry.

PR #123, opened 2015-12-16
Fixed some DocBlocks that had missing sections and wrong types.

diegocaprioli:
These are just tweaks in the DocBlocks, not highly important. 

I got myself confused when I received a "\Symfony\Component\HttpFoundation\RedirectResponse" instead of "\Illuminate\Http\RedirectResponse" from the GoogleProvider::redirect() method. Actually that was right... only the DocBlock was wrong.

### DIFF
--- a/src/One/AbstractProvider.php
+++ b/src/One/AbstractProvider.php
@@ -40,7 +40,7 @@ abstract class AbstractProvider implements ProviderContract
     /**
      * Redirect the user to the authentication page for the provider.
      *
-     * @return RedirectResponse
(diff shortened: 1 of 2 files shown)


src/One/AbstractProvider.php
```diff
@@ -54,6 +54,7 @@ public function redirect()
     /**
      * Get the User instance for the authenticated user.
      *
+     * @throws \InvalidArgumentException
```
GrahamCampbell: not in the correct order

PR #128, opened 2016-01-08
Allow retrieval of raw user data from AbstractUser

davzie:
There is currently no nice way of retrieving the raw data set by the provider. This should change that.


sajjadrad:
+

PR #132, opened 2016-02-26
Support driver with dynamic config

peppeocchi:
That's probably an edge case, but when working with multiple websites you may want to use a driver instance with dynamic config (E.g. coming from a local database or an api call etc...), the credentials could be different for each website as well as the redirect url.

Example use is:
```php
...
$config = [
    'client_id'     => $current['client_id'],
    'client_secret' => $current['client_secret'],
    'redirect'      => 'https://' . $current['domain'] . '/your-callback-url',
];
...
return Socialite::withDynamic('facebook', $config)->redirect()
```

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -4,13 +4,52 @@ namespace Laravel\Socialite;
 
 use InvalidArgumentException;
 use Illuminate\Support\Manager;
-use Laravel\Socialite\One\TwitterProvider;
-use Laravel\Socialite\One\BitbucketProvider;
-use League\OAuth1\Client\Server\Twitter as TwitterServer;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Holding off on this.

AndyWendt:
@peppeocchi we have the ability to do this in our [socialite providers](https://socialiteproviders.github.io/)

peppeocchi:
@AndyWendt thanks for the tip!

amsanket22:
@AndyWendt I think it doesn't work - its mentioned in the documentation but we get 
image
this error if there is no configuration in services.php file

amsanket22:
Oh it works, even if you are using it dynamically as I mentioned above - only catch is you have to still keep your configuration array in services.php - env values can be blank but that configuration array in services.php file is required

someone needs to update the documentation

PR #133, opened 2016-02-27
fetch user with parameters from facebook

yutaro-ihara:
facebook users can set multi names per languages.
so, i want to set locale when i fetch user info.


taylorotwell:
Going to hold off on this.

AndyWendt:
@yutaro-ihara we are allowing quite a few overrides in our [socialite providers](https://socialiteproviders.github.io/) so we may have something that will work for this.   If not we can create a different Facebook provider that allows something like this.

yutaro-ihara:
i see, thanks.

PR #134, opened 2016-03-01
Added a Pinterest Provider.

JaffParker:


GrahamCampbell:
Sorry, we're not accepting new providers to the core.

JaffParker:
So how is it possible to have new providers added to socialite?

I mean, I can just add them to my fork and keep rebasing from laravel/socialite with every update, but it doesn't seem like the most convenient way. Is there anything better?

GrahamCampbell:
https://github.com/SocialiteProviders

GrahamCampbell:
infact, this package already exists: https://github.com/SocialiteProviders/Pinterest

JaffParker:
Yes, I tried to use that package, but it had problems loading. Somehow it would either not find the Pinterest class or try to define it twice, which would cause it crash either way.

GrahamCampbell:
Please feel free to take that up with then. Please close this.

taylorotwell:
Don't think we want to bring a ton of providers in core because they are tough to maintain honestly. Would encourage you to release this on your own GitHub as an extension, etc.

AndyWendt:
@JaffParker  If it is broken, we would appreciate an issue being created on the repo.  Most issues are that it wasn't setup correctly since there really is nothing to throwing these providers together.

AndyWendt:
@JaffParker seems like the issue may have been fixed: https://github.com/SocialiteProviders/Pinterest/issues/2

PR #136, opened 2016-03-13
Add notes for usage with stateless API.

websanova:


taylorotwell:
Needs actual description of what "stateless API" means in this context, etc.

PR #143, opened 2016-04-29
Expose OAuth2 refresh_token and expires_in when available

asaph:
Previously the OAuth2 refresh token was inaccessible. This PR exposes it in the User object when it is returned by the provider. I've regression tested this change with each of the supported OAuth2 providers: Google, Facebook, GitHub and LinkedIn.

Note: [Google only includes the refresh_token on the first authorization](http://stackoverflow.com/a/10857806/166339). To force Google to include the refresh_token every time (such as you might do while validating this PR!), you can pass a couple of additional parameters as shown below.

```php
public function redirectToGoogle() {
    return Socialite::with('google')
            ->scopes(['profile', 'email'])
            ->with(['access_type' => 'offline', 'approval_prompt' => 'force'])
            ->redirect();
}
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -109,6 +109,8 @@ $user = Socialite::driver('github')->user();
 
 // OAuth Two Providers
 $token = $user->token;
+$refreshToken = $user->refreshToken; // not always provided
+$expiresIn = $user->expiresIn;
 
 // OAuth One Providers
 $token = $user->token;
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -68,6 +68,34 @@ abstract class AbstractProvider implements ProviderContract
     protected $scopeSeparator = ',';
 
     /**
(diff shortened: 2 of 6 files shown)


asaph:
Build had passed on this PR but is now failing after I squashed down the style fixes down to 1 commit and force pushed. Looks like it's passing on PHP 5.5, 5.6 and hhvm but failing on the older PHP 5.4. Was PHP 5.4 not part of the CI test when this PR was first opened? In any case, I'll fix the test that doesn't work under PHP 5.4. Stay tuned.

taylorotwell:
Can you make the change simpler without moving all those strings to properties, etc?

asaph:
@taylorotwell Yes. I can do that. The need for the customization/complexity mostly comes from the Facebook OAuth2 implementation which doesn't return json in its access token response. Instead of generalizing the default access token handling logic, I can hardcode it for json and override it in the Facebook-specific subclass. I'll open a new PR for that work. Sound good?

PR #144, opened 2016-05-03
add microsoft sso provider

jswh:
our company uses the office365 services and I wrote an small program with laravel witch should use the SSO of microsoft.So I extended the socialite. Setting up this pull request to make sure if it is appropriate.

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -27,6 +27,28 @@ class SocialiteManager extends Manager implements Contracts\Factory
      *
      * @return \Laravel\Socialite\Two\AbstractProvider
      */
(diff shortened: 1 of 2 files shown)


Gummibeer:
@jswh I don't think that this will be implemented in this package. But you will/can have luck at the community provider collection: https://socialiteproviders.github.io/
There are a lot of ready2use providers and you cann add your provider.

GrahamCampbell:
Sorry, we're not accepting new providers here. Please close this and follow @Gummibeer's suggestion.

jswh:
Got it. Thanks for suggestion.

PR #155, opened 2016-05-21
Add Instagram

rdehnhardt:
although not getting more drivers, I did anyway! If they want to perform the merge, feel free.


taylorotwell:
We prefer to let the community distribute these... see: 

https://socialiteproviders.github.io/

rdehnhardt:
I did not know this package!

It really is better to add driver for driver!

Thanks

PR #156, opened 2016-05-25
Code request parameters merging

hiendv:
Custom parameters should be overridden by required parameters of the code request.
In my situation, setting a custom parameter `state` with `with` resulted in inconsistency between states.
I was thinking about checking existed state before `Str::rand()` a new one but it makes sense to use these prepared fields.

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -175,7 +175,7 @@ abstract class AbstractProvider implements ProviderContract
             $fields['state'] = $state;
         }
 
-        return array_merge($fields, $this->parameters);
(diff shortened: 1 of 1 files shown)


hiendv:
Anyway, injecting `Illuminate\Support\Str` as a dependency would be great. I find it a little confusing to write tests. Thank you.

taylorotwell:
I don't really agree that custom options should be overrideen. Just don't pass a conflicting one.

hiendv:
From my point of view, I don't expect my custom parameters to change because _**they shouldn't be overridden**_ but they do eventually. Overriding custom parameters is evil but in this specific case, it is not.

PR #161, opened 2016-05-30
Added setRedirectUrl method for dynamic URL redirection

KerryJones:
I am using the same configuration on different parts of an application and needed it to be redirected to different locations. I checked forums and it seems this feature was requested elsewhere as well:
http://stackoverflow.com/questions/30487429/laravel-5-and-socialite-new-redirect-after-login

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -350,7 +350,7 @@ abstract class AbstractProvider implements ProviderContract
      * @param  string  $url
      * @return $this
      */
-    public function setRedirectUrl($url)
+    public function redirectUrl($url)
     {
         $this->redirectUrl = $url;
 
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -343,6 +343,19 @@ abstract class AbstractProvider implements ProviderContract
 
         return $this;
     }
+	
+    /**
+     * Set the redirect url after initialization.
+     *
+     * @param  string  $url
+     * @return $this
+     */
+    public function setRedirectUrl($url)
+    {
+        $this->redirectUrl = $url;
+
+        return $this;
+    }
 
     /**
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I don't think this should be allowed. It makes more sense to have it immutable.

KerryJones:
Why should it be immutable? I gave a very plausible scenario (admin. vs main website) where it needs to be a different URL for each scenario. There is no other way to currently change the configuration or redirection.

taylorotwell:
I would just call it `redirectUrl`. Drop the `set`.

KerryJones:
Updated

himinat:
Is there any way to use `'facebook' => [
       ...,
        'redirect' => route('routeName', ['params']),`

GrahamCampbell:
> Why should it be immutable? I gave a very plausible scenario (admin. vs main website)

Should be implemented with two different objects.

KerryJones:
Graham, do you have an example?

Taylor -- let me know if there is anything else you want me to do.

claudiomedina:
Can we have this method for Oauth1 to?

paulocastellano:
@taylorotwell and @KerryJones

The **redirectUrl** still working?

Because if I run:

```
return Socialite::driver('google')
            ->redirectUrl("http://localhost:8000/p/auth/login/google/callback")
            ->redirect();
```

I got error:

```
Client error: `POST https://www.googleapis.com/oauth2/v4/token` resulted in a `400 Bad Request` response: { "error": "redirect_uri_mismatch", "error_description": "Bad Request" }
```

oza75:
@paulocastellano You should also set the redirect URL on the callback route when you retrieve user data
```php
$user = Socialite::driver('google')
           ->redirectUrl("http://localhost:8000/p/auth/login/google/callback")
            ->user();
```

PR #169, opened 2016-06-04
Return \Illuminate\Session\Store

lucasmichot:
Return the expected `\Illuminate\Session\Store` instead of a `\Illuminate\Session\SessionInterface`
See calls to `get`, `set` and `pull`


morloderex:
You'll still get a \Illuminate\Session\Store instance because the interface you are refering to is bound to the same class :)

lucasmichot:
Then what happens when you bind another implementation @morloderex ?

PR #175, opened 2016-07-08
Add age_range field (Facebook)

nickknissen:
Add "age_range" field to get user's facebook [age_range](https://developers.facebook.com/docs/graph-api/reference/age-range/)


taylorotwell:
Is it possible to make a separate call to get this in your own application? Socialite is mainly geared for basic social login functionality.

PR #178, opened 2016-08-12
Add a link to Socialite Providers

misenhower:
I didn't realize this existed at first, so a link might be helpful for people looking to integrate with other platforms. This might also cut down on the number of PRs you get for adding new providers :)


asantibanez:
I second that. Found SocialiteProviders yesterday when wanting to integrate Dropbox.

PR #180, opened 2016-08-28
Update profileUrl (Facebook Provider)

J2TeaM:


taylorotwell:
Holding off on this. Would rather just use the URL given by Facebook.

PR #181, opened 2016-08-31
Added Macroable trait to SocialiteManager

benswinburne:
Example use case such that a package service provider registers a routes method akin to those used by `Auth` etc.

    SocialiteManager::macro('routes', function () {
        app('router')->get('example', function() {
            return 'macroed route';
        });
    });

    Socialte::routes();


GrahamCampbell:
I'm :-1: here. Macros shouldn't be abused. They have a significant performance impact compared to just writing proper code.

taylorotwell:
See above.

PR #182, opened 2016-09-06
Avatar Type added for Facebook

imrealashu:
Now user can specify avatar types i.e., `small`, `normal`, `album`, `large`, `square` using `setAvatarType('square')`.


taylorotwell:
Going to just stick with the convention for now.

PR #184, opened 2016-09-20
Allow overriding the session state string.

geodeveloper:
In some scenarios we could override the session state string with another type of encryption. For example an use case will be to store encrypted information into the session state.


geodeveloper:
@taylorotwell I noticed that on 3.* version this was removed. Is there any reason why? See my original comment.

PR #190, opened 2016-10-29
Update docs on retrieving user via token

tomhorvat:
Introduced in #126


taylorotwell:
Please follow the formatting of the rest of the documentation. We always use capital letters in titles in every other title in the entire Laravel documentation.

PR #193, opened 2016-11-01
add "get user profile" from `token` and `secret`

Mahmoudz:
The `Two/AbstractProvider.php` has `userFromToken($token)` to retrieve
user profile from a `token`.

While `One/AbstractProvider.php` doesn’t have this feature, I added
`userFromTokenAndSecret($token, $secret)` to retrieve user profile from
the  `token` and `secret`. (Tested with Twitter).

### DIFF
--- a/src/One/AbstractProvider.php
+++ b/src/One/AbstractProvider.php
@@ -4,6 +4,7 @@ namespace Laravel\Socialite\One;
 
 use Illuminate\Http\Request;
 use InvalidArgumentException;
+use League\OAuth1\Client\Credentials\TokenCredentials;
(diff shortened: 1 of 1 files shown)


taylorotwell:
I added the doc block for you that you left out.

taylorotwell:
Thanks.

amanjaswalia:
When i try with facebook or twitter login in windows 7 with IE11 getting NULL value from One/AbstractProvider.php  -> getToken() ->   $temp = $this->request->session()->get('oauth.temp')
Not get value from "oauth.temp".
On upper version of IE , all version of firefox and chrome working fine and get value in "oauth.temp".
Please help.

PR #195, opened 2016-12-03
Some Facebook Users dont have an email and it gives an error

psmchart:
Hi,

        I noticed that some facebook users dont have an email when they signup at facebook and used their mobile number instead. And when they signup to my laravel website using Socialite package(facebook) it gives me an error that say "email not null" etc etc.

        Any help is highly appreciated.

Sincerely,
PSMchart

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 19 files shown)


taylorotwell:
As you can see this is not submitted properly. Has a bunch of extra commits.

PR #197, opened 2017-01-10
Set session save compatible with Laravel 5.4

eusonlito:


taylorotwell:
Tests are failing.

eusonlito:
Ok, put not exists in `Mockery_1_Symfony_Component_HttpFoundation_Session_SessionInterface` but exists in Laravel 5.3 and 5.4

`set` exists in Laravel 5.3 (https://github.com/laravel/framework/blob/5.3/src/Illuminate/Session/Store.php#L394) but not in Laravel 5.4 (https://github.com/laravel/framework/blob/5.4/src/Illuminate/Session/Store.php)

No problem, I will use my fork with Laravel 5.4 projects.

Thanks.

morloderex:
@taylorotwell The problem here is that when you refactored the session component away from symfony session. You renamed the method name from `set` to `put`. In order to make socialite compatible with laravel `5.4` we will have to upgrade this package and its tests to not use symfony sessionInterface but laravels session interface i suppose.

taylorotwell:
Yes, these changes should be made on the master branch for the next major release.

PR #199, opened 2017-01-17
Support the stateless in OAuth one with native session, support to Lumen

ktquez:


taylorotwell:
Laravel never uses $_SESSION.

PR #211, opened 2017-03-05
Add getGender()

sbimochan:


src/Contracts/User.php
```diff
@@ -38,4 +38,11 @@ public function getEmail();
      * @return string
      */
     public function getAvatar();
+
+    /**
+     * Get the gender of the user.
+     *
+     * @return string
+     */
+    public function getGender();
```
crynobone: Is this targeting next major release? cause this potentially a huge breaking change.

sbimochan: Absolutely not. I just edited socialite in project to pull gender and thought other people may need this function too.

GrahamCampbell: > Absolutely not.

It needs to, because adding a new method to an interface is a major breaking change.

taylorotwell:
Breaking change.

PR #212, opened 2017-03-13
[3.0] Allow PHPUnit ~6.0 in composer's dev requirements

jeromegamez:
Allows PHPUnit ^6.0 to be used in projects that use Laravel Socialite.

The class aliases only come into effect when executing the Socialite test suite and will not affect projects having it as a dependency.

:octocat:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require-dev": {
         "mockery/mockery": "~0.9",
-        "phpunit/phpunit": "~4.0|~5.0"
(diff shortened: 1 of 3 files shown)


jeromegamez:
StyleCI doesn't like that I force pushed into my branch to keep the commit history clean, but I fixed the whitespace issues that it told me to fix :).

taylorotwell:
A package's require-dev specifications have no bearing on what your own project can use.

PR #213, opened 2017-03-15
bitbucket requires to have auth at this level

lukepolo:


taylorotwell:
Can anyone else independently confirm this? @themsaid?

themsaid:
I can confirm that it works *without* the changes in this PR.

lukepolo:
Then what am i doing wrong, it says I'm missing the client credentials, only works when adding that line of code there ~

lukepolo:
Client error: `POST https://bitbucket.org/site/oauth2/access_token` resulted in a `400 Bad Request` response: {"error_description": "Client credentials missing; this request needs to be authenticated with the OAuth client id and s (truncated...)


image

taylorotwell:
Closing for now since we can't recreate this issue. If you can recreate it on a fresh, simple Laravel application and link @themsaid to it maybe he could recreate it then?

lukepolo:
Yup will try it.

lukepolo:
https://github.com/lukepolo/BitBucketTest

There is the repository , same issue 

image

rdgout:
I have the same issue as above with a fresh Laravel installation

lukepolo:
@rdgout we may have to open a new issue not sure if this will get noticed~

rdgout:
@themsaid could you verify these findings? It's a little tedious to modify the source of the package when I deploy my application.

lukepolo:
@rdgout @themsaid were you able to solve this without my patch?

rdgout:
@lukepolo Nope, I had to manually change the code in the package when I deployed. Which means that I'm screwed when it's updated without the fix. Maybe we should ping @taylorotwell so he can take another look if @themsaid is busy.

themsaid:
Hey, I'll look into the repository you created now and try to replicate.

themsaid:
I still confirm that it's working, I hit `http://zzzz.ngrok.io/oauth/callback/bitbucket` and then grant access on BitBucket, then get redirected back to `http://zzz.ngrok.io/oauth/callback/bitbucket?state=TT6s2S3S9fFoIJTlv0M1E5OS0p24AlsCxX0QaCd2&code=xFYVQemDez5WA8RU7D` with the user data needed.

lukepolo:
did you give the scopes as following 


image

themsaid:
yes

themsaid:
I guess you may need to contact BitBucket and report this behaviour.

lukepolo:
Can i send you my client id and secret via another method , see if that works ? Im just trying to figure out why it wont work for me while it works for you perfectly

lukepolo:
Ok will report back here after i contact them  ~

themsaid:
@lukepolo ok please ping me when you do, or say hello on twitter @themsaid :)

themsaid:
If apply the changes described by @dtao I get:

```
Client error: `GET https://api.bitbucket.org/2.0/user?access_token=` resulted in a `401 Unauthorized` response:
{"type": "error", "error": {"message": "Access token expired. Use your refresh token to obtain a new access token."}}
```
Without passing `auth` the code just works, Also merging @lukepolo PR locally causes guzzle to break since you're passing the username only not the password for the auth.

dtao:
@themsaid Do you know what version of the code you're using? I am a little confused because there appear to be 3 branches: master, 2.0, and 3.0; and the problem I described only exists in the 3.0 branch (where, incidentally, you can see that `getAccessToken` literally [only appears in one place, BitbucketProvider.php](https://github.com/laravel/socialite/search?utf8=%E2%9C%93&q=getAccessToken&type=)).

themsaid:
@dtao I'm on 3.0, I think the `getAccessToken()` can be removed since the `getAccessTokenResponse()` is being used and it works. Can you please share that part in the docs where it says auth headers should be passed?

dtao:
@themsaid The portion of the RFC I referred @lukepolo to was [section 3.2.1](https://tools.ietf.org/html/rfc6749#section-3.2.1):

> Confidential clients or other clients issued client credentials MUST
   authenticate with the authorization server as described in
   Section 2.3 when making requests to the token endpoint.

That said, on closer inspection I see that the default implementation in the `AbstractProvider` class _does_ pass `client_id` and `client_secret` [here](https://github.com/laravel/socialite/blob/3.0/src/Two/AbstractProvider.php#L267-L273), which should work as an authentication mechanism. (Basically Bitbucket will return a 400 if you haven't provided _either_ an authorization header or these values.)

@lukepolo Is it possible these values are not set properly for you? Once again, I'm not very familiar with this library so I don't know if that question makes sense; but if that's the case then that would explain why it is working for other contributors to this thread but not for you.

lukepolo:
Ok so with 3.0.3 it does not work , but with 3.0.4 it does just tested this. My repo was using 3.0.3 so not sure why it didnt trigger for you earlier ~

lukepolo:
https://github.com/laravel/socialite/commit/82882dced2ad9d25f7fa8c3aca77b67b8ebf02c7

Is the change that made it work

rdgout:
I can confirm it works out-of-the-box now rather than the problem we were encountering before.

PR #214, opened 2017-03-27
Fixing access token parsing

tomsisk:
Specifying version when requesting access token. 

v2.2 of Facebook API (last version to use query strings here) is no longer available, starting today.

https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow

> Note: From v2.3 onward this endpoint will return a proper JSON response. If your call doesn't specify a version it will default to the oldest available version.

### DIFF
--- a/src/Two/FacebookProvider.php
+++ b/src/Two/FacebookProvider.php
@@ -62,7 +62,7 @@ class FacebookProvider extends AbstractProvider implements ProviderInterface
      */
     protected function getTokenUrl()
     {
-        return $this->graphUrl.'/oauth/access_token';
+        return $this->graphUrl.'/'.$this->version.'/oauth/access_token';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


RandyBooth:
Came to look for this problem with v2.2 end of life today. Hopefully get this update soon. Thank you tomsisk

alberto1el:
this worked for me thanks @tomsisk

diogolast:
what a relief, thanks @tomsisk

ceejayoz:
Thanks for the fix. We're on 2.0 but this issue saved us what would probably have been a lot of debugging.

anishjoshi:
Thanks @tomsisk :) worked on my end.

andreikainer:
Cheers. This just saved our live application! :-)

truongphitam:
Thanks for the fix. But is loops.

DoanTran:
@andreikainer Me too! Thanks @tomsisk

jiachiun:
Thanks for the fix! +1

coreation:
@andreikainer same here, thanks a bunch @tomsisk , I had no clue the API was going to be deprecated, any idea on how to keep track of API changes in FB?

tomsisk:
@coreation https://developers.facebook.com/docs/apps/changelog gives you an overview. Also see https://developers.facebook.com/tools/api_versioning/

Your developer console will also have alerts if you're using soon-to-be expired versions. Now that we're specifying the version when retrieving an access token, we should be good for the foreseeable future.

fazlurr:
+1 Big thanks..

PR #216, opened 2017-03-28
Change picture type to square

gizburdt:
In Facebook it is possible to set a vertical or horizontal image. When using the type=normal, and setting the height and width to 32x32, the image will be squeezed. Square already returns an image width same height.


taylorotwell:
Use a different cropping mechanism. Most image libraries can crop to center without squeezing, etc.

PR #219, opened 2017-04-05
Check if the authorization code is present

nesk:
OAuth 1 providers check if the OAuth verifier is present but OAuth 2 providers don't.

I've added the same verification for OAuth2 providers. This allows to catch all the errors about malformed requests with a simple `try…catch`:

```php
try {
    $user = Socialite::driver('any-driver')->user();
} catch (InvalidArgumentException $e) {
    // The authorization request is missing either an OAuth verifier or an authorization code.
    // This can happen when the user cancels the authentication on the remote website.
}
```

This will not break the third-party providers in the [Socialite Providers](https://socialiteproviders.github.io/) project since [they rewrote the `user` method](https://github.com/SocialiteProviders/Manager/blob/master/src/OAuth2/AbstractProvider.php#L29), so you can merge with confidence.

However, once this pull request will be merged, I will make a new pull request to the [Socialite Providers](https://socialiteproviders.github.io/) project so they handle this verification too.

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -6,6 +6,7 @@ use GuzzleHttp\Client;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Illuminate\Http\Request;
+use InvalidArgumentException;
 use GuzzleHttp\ClientInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Laravel\Socialite\Contracts\Provider as ProviderContract;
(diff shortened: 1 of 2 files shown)


nesk:
Just rebased my commit to add a test case

taylorotwell:
What is the current behavior of the user() method if the code is not present?

nesk:
We get the following exception:

```
GuzzleHttp\Exception\ClientException: Client error: `POST https://graph.facebook.com/v2.8/oauth/access_token` resulted in a `400 Bad Request` response:
{"error":{"message":"Missing authorization code","type":"OAuthException","code":1,"fbtrace_id":"**A trace ID**"}}
 in /path/to/the/project/vendor/guzzlehttp/guzzle/src/Exception/RequestException.php:111
```

houres:
exactly, and this is really frustrating.

taylorotwell:
OK then I can't change this on a minor release because people could already be catching this error and handling it, and if I change the exception type their code will break.

nesk:
Will you merge this PR for a potential v4?

PR #226, opened 2017-05-06
"Unable to verify access token" error while using LinkedIn mobile SDK tokens and getUserByToken method

vkovic:
Make mobile SDK tokens work with LinkedIn driver while using getUserByToken method.

This is working with Facebook driver (did not try others), so I thought it would be great to work with LinkedIn driver also.

### DIFF
--- a/src/Two/LinkedInProvider.php
+++ b/src/Two/LinkedInProvider.php
@@ -69,6 +69,7 @@ class LinkedInProvider extends AbstractProvider implements ProviderInterface
 
         $response = $this->getHttpClient()->get($url, [
             'headers' => [
+                'x-li-src' => 'msdk',
                 'x-li-format' => 'json',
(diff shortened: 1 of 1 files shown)


vkovic:
Just checked, now regular tokens are not working, just those issued by mobile sdk, so i'm closing  this pull request.

Furthermore: https://developer.linkedin.com/docs/android-sdk-auth

>Mobile vs. server-side access tokens
>
>It is important to note that access tokens that are acquired via the Mobile SDK are only useable with the Mobile SDK, and cannot be used to make server-side REST API calls.
>
>Similarly, access tokens that you already have stored from your users that authenticated using a server-side REST API call will not work with the Mobile SDK.

Thus with header "x-li-scr" header set to "msdk" this actually works, but it's for another topic ...

PR #227, opened 2017-05-15
Using the value helper so a closure can be provided

robinvdvleuten:
As I read in PR #121, there shouldn't be extra dependencies added to the socialite packages. But it become very handy to generate an url automatically by route name in a multi domain / stages setup. So I've added the `value` helper so you can call any "route generation" method within a closure;

```php
'google' => [
        'client_id' => env('GOOGLE_CLIENT_ID'),
        'client_secret' => env('GOOGLE_CLIENT_SECRET'),
        'redirect' => function () {
                return route('name_of_the_callback_route');
        }
],
```

Hopefully you'll like the idea!

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -106,7 +106,7 @@ class SocialiteManager extends Manager implements Contracts\Factory
     {
         return new $provider(
             $this->app['request'], $config['client_id'],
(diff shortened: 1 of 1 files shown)


m1guelpf:
@taylorotwell Could you tag a new version so we can use this?

PR #232, opened 2017-05-24
Use Illuminate's RedirectResponse instead of Symfony's

TimWolla:
Illuminate extends Symfony's RedirectResponse with ResponseTrait, which
introduces the `header()` method. By not using Illuminate's
RedirectResponse Socialite breaks middlewares that try to add headers to
every response (e.g. no-cache headers).

### DIFF
--- a/src/One/AbstractProvider.php
+++ b/src/One/AbstractProvider.php
@@ -4,9 +4,9 @@ namespace Laravel\Socialite\One;
 
 use Illuminate\Http\Request;
 use InvalidArgumentException;
-use League\OAuth1\Client\Credentials\TokenCredentials;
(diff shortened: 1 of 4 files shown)


TimWolla:
I'll happily change the StyleCI CI failures, but the suggested changes in the Details link look strange to me. Why does it separate the different `Illuminate` imports in Two/AbstractProvider.php? Can anyone advise?

GrahamCampbell:
Laravel's code style is to order imports by length, not alphabetically.

TimWolla:
> Laravel's code style is to order imports by length, not alphabetically.

Thanks for your clarification. I just amended my commit to fix style and force pushed.

PR #233, opened 2017-05-25
Change $scopeSeparator to a space in BitbucketProvider

Daedalus11069:
In Bitbucket's implementation of OAuth2, the space character is the delimiter for scopes.

When trying to request multiple scopes, the request fails with the callback url having `error_description` appended to it with the value of 'Unknown scope: ', and then a comma-delimited list of scopes.  After that the listed `error` variable has a value of 'invalid_scope'.


Daedalus11069:
I'm not sure how I messed up on the style, and I'm sorry for that failure.  I'll fix it if someone could tell me how to.

PR #237, opened 2017-06-30
Added branch reference to composer command

Dreanmer:
This will help some people like me :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ In addition to typical, form based authentication, Laravel also provides a simpl
 
 To get started with Socialite, use Composer to add the package to your project's dependencies:
 
-    composer require laravel/socialite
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎  Composer should decide what to install, and v2 is not the latest, and is also not supported anymore.

raul3k:
Well, don't know if it's related but when I try to install this package, the version 3.0, composer shows that it is downloading the version 3.0, shows in composer.lock the version is 3.0 but the code is the version 2.0.

Dreanmer:
@GrahamCampbell yeah sorry i made it on the wrong branch, its because i'm using laravel 5.3.x so i can't use latest version (as the readme says) also it gave me an error if i try:
image
so i'll remade the PR for the correct branch

PR #240, opened 2017-07-24
Added branch reference to composer command

Dreanmer:
###the problem:
when you try to require socialite into laravel 5.3 or bellow you will have the following error:
image
###the solution:
as the 3.0 readme says: install 2.0 version, but on the 2.0 readme, the composer require command will install the version 3.0, so, i specified the ^2 version on the composer command.


taylorotwell:
Shouldn't be necessary.

PR #243, opened 2017-08-22
Update readme

luiz-brandao:
Add information about the new relative path option.

PR #244, opened 2017-08-22
Update readme

luiz-brandao:
Add information about the new relative path option. This fixes the typo found in the last pull request #243.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,9 @@ You will also need to add credentials for the OAuth services your application ut
     'redirect' => 'http://your-callback-url',
 ],
 ```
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not entirely sure any of this was ever necessary since you always could have done:

'redirect' => url('/relative/path'),

luiz-brandao:
Oh right, so I probably missed that. I was trying to use `route()` but I was having a hard time since routes didn't seem to be available at the config time for some reason. Anyway, thanks and revert it if you feel it's the best thing to do.

PR #245, opened 2017-08-22
Support Auto-Discovery

bambamboole:
I think this is the only needed change to support auto discovery of the package .


matijakovacevic:
shouldn't README also be updated noting that provider and alias doesn't need to be manually inserted into config?

PR #250, opened 2017-11-02
[3.0] Updated dev dependencies

gabriel-caruso:
Updated `mockery/mockery` to `~1.0` and `phpunit/phpunit` to `~6.0`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,8 @@
         "league/oauth1-client": "~1.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


taylorotwell:
Tests fail.

gabriel-caruso:
@taylorotwell The tests only failed because of PHPUnit 6.0 [requires PHP 7.0](https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L25). Should we drop support to `5.6`?

PR #253, opened 2017-11-06
Fixed missing import statement

JapSeyz:
Error given at L155 because Str is not imported.


RomanGorbatko:
@JapSeyz 

Have this error too.
Waiting for update.

PR #254, opened 2017-11-14
[3.0] Documentation cleanup

paulofreitas:
Now that Laravel Socialite's was documented both in `5.4` and `5.5` branches of `laravel/docs`, it should be fine cleaning it up here so further package documentation updates get added into the right place. :+1:


WadeShuler:
Since this repo doesn't support `issues`, I am simply replying here as it appears to be the most appropriate place I can find (so far) as it's related to docs.

The docs say to use:

    $user = Socialite::driver('twitter')->userFromToken($token);

However I am getting:

    Call to undefined method Laravel\Socialite\One\TwitterProvider::userFromToken()

I inspected `TwitterProvider` and that function doesn't exist. So I looked at what it extends, `AbstractProvider`, which also doesn't have the function `userFromToken`. It does have the function `userFromTokenAndSecret`.

So I am confused at why the docs, guide, and tutorials I can find, all say to use `useFromToken` and I can't find anything on `userFromTokenAndSecret`.

I can't for the life of me get `userFromToken` to work, but just tried `userFromTokenAndSecret` and it works.

Is this documented or do the main installation instructions on your site and the readme of this repo, still need to be updated?

paulofreitas:
@WadeShuler You're right, for `TwitterProvider` you should be using the `userFromTokenAndSecret()` method. I guess it wasn't documented yet because currently only the `TwitterProvider` is using this OAuth1 interface, all other providers are based on OAuth2 interface where `userFromToken()` comes. 👍 

So yes, the documentation should be updated to mention this other method. Note that Socialite documentation is being moved to [laravel/docs](https://github.com/laravel/docs) repository. Can you please make a pull request there? 😃

WadeShuler:
Thank you for your reply. I've been coding for a while, but fairly new to the Laravel universe. So I appreciate your polite reply.

Is the Twitter module going to be upgraded to OAuth2 in the near future? Just wondering, since it's the only one that isn't. :)

I will try and make time soon (and remember) to PR the doc.

Thanks again!

PR #261, opened 2018-05-11
Bump Graph API version to v3.0

kamui545:


kamui545:
The `link` field has been deprecated and soon (1st august) will require the `user_link` scope for **every versions**.

**HOWEVER**, the code is already designed in a way that this field is **optional**.

```php
    protected function mapUserToObject(array $user)
    {
        $avatarUrl = $this->graphUrl.'/'.$this->version.'/'.$user['id'].'/picture';

        return (new User)->setRaw($user)->map([
            'id' => $user['id'], 'nickname' => null, 'name' => isset($user['name']) ? $user['name'] : null,
            'email' => isset($user['email']) ? $user['email'] : null, 'avatar' => $avatarUrl.'?type=normal',
            'avatar_original' => $avatarUrl.'?width=1920',
            'profileUrl' => isset($user['link']) ? $user['link'] : null, // here
        ]);
    }
```

It would be nice to have this merged, breaking changes to Facebook API (no matter the versions) will be in effect 1st august, and apps needs to be reviewed again before this 90 days deadline...

> New Features, Changes, and Deprecations only affect this version. 90-Day Breaking Changes affect all versions.

https://developers.facebook.com/docs/graph-api/changelog/version3.0

Tagging this as v3.1 might still be better, even if it should work just fine.

PR #262, opened 2018-05-23
Change docblock return tag to \Illuminate\Http\Response

jrhenderson1988:


GrahamCampbell:
I think the original version was already fine. The contract just says that implementers must return at least the symfony redirect response.

PR #264, opened 2018-06-09
Get authorization URL to use with API and SPA

emtudo:


emtudo:
Well! In fact is not necessary, beacuse I can use:

        $server = new Server(['identifier' => 'id', 'secret' => 'secret']);
        $server->getAuthorizationUrl($server->getTemporaryCredentials())

PR #270, opened 2018-06-17
[3.0] Extract post key resolution.

lucasmichot:
Extract post key resolution.


taylorotwell:
I think it's OK as is.

PR #273, opened 2018-06-18
Bump Graph API version to v3.0

johankrijt:


taylorotwell:
This isn't the current release of Socialite.

johankrijt:
Yes, but it is the release used by Laravel 5.3 or below. Facebook is pushing API version 3.0 with all new Facebook applications, so would be a shame if anyone using Laravel 5.3 or lower couldn't use Socialite out of the box.

PR #274, opened 2018-07-09
Use Google Sign-in API instead of G+ for User Info

eheaton:
I'm not sure if this driver was originally designed explicitly for Google+, or just regular Google Sign-In. 

The below PR suggests that we remove the dependency of the G+ API on the Google Apps. The regular `/userinfo/v2/me` API provides the information socialite needs to get without any additional API's.

One thing to note about relying on the G+ API call - it requires the user to have set up / enabled their Google+ Profile, which is no longer done automatically on newly created Google accounts, and disabled on some G-Suite accounts.

### DIFF
--- a/src/Two/GoogleProvider.php
+++ b/src/Two/GoogleProvider.php
@@ -58,10 +58,7 @@ class GoogleProvider extends AbstractProvider implements ProviderInterface
      */
     protected function getUserByToken($token)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Feels like something we shouldn't be changing on a point release?

eheaton:
No worries, that makes sense. It'd be good to have it considered for the backlog on the next major release.

PR #281, opened 2018-10-03
update to use google OAuth2

tanleekent:
you can use this api to get user info without needed to add google+ api in the google console project


taylorotwell:
Honestly I need feedback from other people using Socialite with Google on if this is a good change or not. I don't use that driver.

PR #283, opened 2018-10-09
changed API url to not use google+, updated mapping of user object

wgmv:
In the previous version the developer needed to activate the google+ api in googles dev console. (not activating it results in 

```Client error: `GET https://www.googleapis.com/plus/v1/people/me?prettyPrint=false` resulted in a `403 Forbidden` response: {"error":{"errors":[{"domain":"usageLimits","reason":"accessNotConfigured","message":"Access Not Configured. Google+ API ```

With this api url we can get the email, name and avatar without google plus. I heard it is shutting down - so maybe this is useful.

Thanks and regards.


narwy:
Looks fine to me

PR #291, opened 2018-10-27
Gets all emails in addition to the main email for GithubProvider.php. 

iliyaZelenko:
Perhaps this will interfere with compatibility, if this variant does not fit then see my second PR.


taylorotwell:
No plan to add right now.

PR #292, opened 2018-10-27
Gets all emails in addition to the main email for GithubProvider.php. Variant 2

iliyaZelenko:
This is the second variant.

[Here is the first](https://github.com/laravel/socialite/pull/291).


taylorotwell:
No plans to add right now.

PR #298, opened 2018-11-10
[3.0 ]Use modern phpunit TestCase namespace.

lucasmichot:
]Use modern `phpunit` TestCase namespace.


driesvints:
@lucasmichot I'm a bit confused how this can work. We require `~4.0|~5.0` in our composer.json and if you look at the Travis build it installs 5.7.27 which doesn't has namespaces yet. They were only added in 6.0. I'm not really sure what's going on and how this could work.

PR #303, opened 2018-11-26
[3.0] Added stateless to twitter

ErFUN-KH:
Now it's possible to use Twitter without session.

### DIFF
--- a/src/One/AbstractProvider.php
+++ b/src/One/AbstractProvider.php
@@ -2,8 +2,8 @@
 
 namespace Laravel\Socialite\One;
 
-use Illuminate\Http\Request;
 use Illuminate\Support\Str;
+use Illuminate\Http\Request;
 use InvalidArgumentException;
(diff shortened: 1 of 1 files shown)


driesvints:
@ErFUN-KH I think it's best that you add more context why you're sending this in, link to examples from the Twitter api, etc. More context helps us explain why this could be useful to others.

ErFUN-KH:
@driesvints Twitter using OAuth 1.0 and others like Google and Facebook, etc using OAuth 2.0, so stateless method is available for OAuth 2.0 but in OAuth 1.0 you must use the session. I use cache instead of Session because when people using Lumen they can't use the session and it doesn't exist at all.
I'm developing a RestFul service with JWT and I need to a stateless method.
is it enough?

taylorotwell:
Holding off on this for now. Feel free to release a custom driver package like `stateless-twitter` or something.

PR #313, opened 2018-12-19
set correct property

jsharma44:


taylorotwell:
Please remove unrelated white space changes and give reference for why these changes are needed. Are they breaking? etc.

Issue #314, opened 2018-12-19
Issue with Google+ Provider


bocahptualang:
Error when authenticate with google+ provider (Socialite version 4.0.0). 
I tried downgrade socialite version to 3.1.1 and it is working. 

![image](https://user-images.githubusercontent.com/9653331/50205537-2bbafe00-039b-11e9-8829-224155f0adfa.png)

`GoogleProvider.php`
![image](https://user-images.githubusercontent.com/9653331/50205464-e5fe3580-039a-11e9-8a66-cb3272749cad.png)

Anyway, is there any way to override mapping function from response we get from provider to user object? it would be nice if got one.

rjvandoesburg:
I too am experiencing this problem with a newly created token. On another website (different token) I am still receiving `displayName`.

Attached 2 different responses:
[old.txt](https://github.com/laravel/socialite/files/2694966/old.txt)
The old one is from another implementation

[new.txt](https://github.com/laravel/socialite/files/2694965/new.txt)
This is the response I'm getting from a newly generated key.

As they differ so much I think I'm doing something different so I'll keep looking into it.

Cheers,

NicolaeCasir:
Good job man. I just wanted to make PR... With email its OK.. but with name, its a problem... i've edited GoogleProvider in vendor but im waiting for merge!

rjvandoesburg:
Merging the PR will break my other implementations so more investigation is needed or the PR need to be updated to accommodate both responses

driesvints:
Just merged the PR. I'll tag a new release in a bit.

PR #316, opened 2018-12-20
Fix mapping Google Userinfo to User Object

joshbrown:
It seems that Google's response when fetching the user info has changed.

Updated to reflect those changes. I could not find any change log on Google's side for reference.

#314 has some more info along with the specific error.

Google Returns the following for reference:
```
[
    "user" => [
        "id" => "",
        "email" => "",
        "verified_email" => "",
        "name" => "",
        "given_name" => "",
        "family_name" => "",
        "link" => "",
        "picture" => "",
        "locale" => ""
    ]
]
```

### DIFF
--- a/src/Two/GoogleProvider.php
+++ b/src/Two/GoogleProvider.php
@@ -76,13 +76,13 @@ class GoogleProvider extends AbstractProvider implements ProviderInterface
      */
     protected function mapUserToObject(array $user)
     {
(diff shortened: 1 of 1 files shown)


driesvints:
Looks like the fix from https://github.com/laravel/socialite/pull/283/files got reverted during a merge. Thanks for sending this in. 

@rjvandoesburg don't think that's necessary as they key isn't present anymore.

PR #325, opened 2019-01-04
Fix LinkedInProvider for new API

jeromeongcol:


driesvints:
Please upgrade to 4.0 for using the new api. 3.0 offers support for the old implementation.

jeromeongcol:
4.0 is not backwards compatible to laravel 5.5 apparently. but thank you

PR #326, opened 2019-01-08
[4.0] Fix return type inconsistency

antonkomarev:
Changed return type to `\Illuminate\Http\RedirectResponse` because in reality it's returned in `AbstractProvider`.
I understand that it's subclass and satisfies this contract, but then it's inconsistent return types, because we are using Laravel specific `\Illuminate\Http\Request` instead `Symfony\Component\HttpFoundation\Request` parent class.

### DIFF
--- a/src/Contracts/Provider.php
+++ b/src/Contracts/Provider.php
@@ -7,7 +7,7 @@ interface Provider
     /**
      * Redirect the user to the authentication page for the provider.
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @return \Illuminate\Http\RedirectResponse
      */
     public function redirect();
 
--- a/src/Two/ProviderInterface.php
+++ b/src/Two/ProviderInterface.php
@@ -7,7 +7,7 @@ interface ProviderInterface
     /**
      * Redirect the user to the authentication page for the provider.
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @return \Illuminate\Http\RedirectResponse
      */
     public function redirect();
 


driesvints:
But that's the whole point of a contract... The contract is satisfied. It doesn't matter which concrete is returned.

antonkomarev:
You are right. And I understand that it's correct return type if this contract will be used in Symfony application. But as I understand it's focused only on Laravel and I don't really sure if we need to use Symfony's types in contract.

I should clarify why did I noticed it. I've created Laravel package which using Socialite under the hood.

```php
private function redirectToAuthorization(
    string $provider
) : \Illuminate\Http\RedirectResponse
{
    // ...some magic

    return $socialite->redirect();
}
```
I were wanted to use `\Illuminate\Http\RedirectResponse` return type, but it was highlighted that's it's wrong and I should return `\Symfony\Component\HttpFoundation\RedirectResponse`.

Okay, I've changed it to satisfy the contract. But then I've understood that I need to change all the return types of all the methods which are using `redirectToAuthorization` method and so on it's proxied to my controllers and I have to use Symfony's types in my Laravel app.

antonkomarev:
As I remember in earlier Laravel versions we've got Symfony return types in Form Requests and somewhere else, but then they were replaced by Laravel specific types. That's why I've thought it might be a possible change.

driesvints:
>  I don't really sure if we need to use Symfony's types in contract.

I don't see why not. What's being returned is important for the internals. What's being used in user-land (the controller) doesn't matters in this case.

You can safely change your return type of your method to the Symfony one because that's what's being returned by the contract. Laravel can handle those.

I do see your point about the return types in the `AbstractProvider`. I think those might be better off being the same as the contract. 

These should be native return types anyway to best off but that would be a breaking change.

driesvints:
I do have to admit that because this being a Laravel only package this PR does makes sense a bit. So in the end if it were to be merged it wouldn't be too bad. I don't see this package being used on a standalone basis.

GrahamCampbell:
👎 for merging. The existing return types are correct.

PR #327, opened 2019-01-10
[PoC] Config drivers on the fly

antonkomarev:
It's just proof of concept to open discussion how better to implement it. 
This PR not include tests because I'm not sure that there are plans to add this feature.

I need to change driver configuration on the fly. Now to do it we need:

1. Create `CustomSocialiteManager` which extends `SocialiteManager`
2. In `CustomSocialiteManager` overwrite all these methods: `createGithubDriver`, `createFacebookDriver`, `createGoogleDriver`, `createLinkedinDriver`, `createBitbucketDriver`, `createGitlabDriver`, `createTwitterDriver`.
3. Bind `CustomSocialiteManager` to the `Laravel\Socialite\Contracts\Factory`.

### Solution 1

This is how with this PR we will be able to do it.

```php
/**
 * @param \Laravel\Socialite\Contracts\Factory $socialite
 */
public function __construct(Factory $socialite) {
    $this->socialite = $socialite;
}

public function redirect()
{
    $socialite = $this->socialite->configure([
        'client_id' => 'CLIENT_ID',
        'client_secret' => 'CLIENT_SECRET',
        'redirect' => 'REDIRECT_URL',
    ])->driver('github')->redirect();
}
```

### Solution 2

This is another approach which in my point of view better, but more complex.

```php
/**
 * @param \Laravel\Socialite\Contracts\Factory $socialite
 */
public function __construct(Factory $socialite) {
    $this->socialite = $socialite;
}

public function redirect()
{
    $config = new Config('CLIENT_ID', 'CLIENT_SECRET', 'REDIRECT_URL', [
        'additional' => 'param',
    ]);

    $socialite = $this->socialite
        ->configure($config)
        ->driver('github')
        ->redirect();
}
```

### DIFF
--- a/src/Contracts/Factory.php
+++ b/src/Contracts/Factory.php
@@ -11,4 +11,12 @@ interface Factory
      * @return \Laravel\Socialite\Contracts\Provider
      */
     public function driver($driver = null);
+
+    /**
+     * Configure driver.
+     *
+     * @param array $config
+     * @return $this
+     */
+    public function configure(array $config);
 }
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -17,6 +17,8 @@ use League\OAuth1\Client\Server\Twitter as TwitterServer;
 
 class SocialiteManager extends Manager implements Contracts\Factory
 {
+    private $config;
+
     /**
      * Get a driver instance.
      *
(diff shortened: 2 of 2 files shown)


antonkomarev:
If the second solution will be implemented then [SocialiteProviders/Manager](https://github.com/SocialiteProviders/Manager/blob/master/src/ConfigTrait.php) trait could be replaced with native function call.

driesvints:
Curious: *why* do you need to change your configuration on the fly?

antonkomarev:
@driesvints because I have multi-tenant application and each tenant could have own social providers applications configuration.

driesvints:
Hmm, can you explain a bit further? How would this exactly work?

driesvints:
I see. Yeah that could be useful. Although I'm not sure if this PR is the correct solution. Maybe a more generic solution with the current providers would be better. Maybe the `Manager` class isn't the best way to go in that case. But tbh we're now dangerously diverting from the path where this package is intended for and I'm not really sure that this is the path we want to take.

Maybe your use case is something for a 3rd party package?

antonkomarev:
3rd party package already exists: https://github.com/SocialiteProviders/Manager
It's just extension of Socialite and working good, but it doesn't have configuration for drivers which are implemented in stock Socialite. You have to create those drivers yourself and store them in your app.
You think that configuration should be a part of each provider?

driesvints:
I think this is best solved in the SocialiteProviders package itself if the base functionality is there already. Maybe overwrite the base providers there.

taylorotwell:
No plans to change this right now.

m1guelpf:
SocialiteProviders mantainer here, feel free to send some PRs :wink:

PR #329, opened 2019-01-21
LinkedIn provider tweaks

Crypton33:
The name is now displayed correctly for users outside of the USA (I had the pl_PL locale, so the "firstName.localized.en_US" returned an empty string).

Also exposed the first_name and last_name fields which is useful in the forms and populating user table columns.


driesvints:
Released 4.0.3

PR #336, opened 2019-02-04
[4.0] Google avatar is default parameter

No148:
Bring back is_default parameter for avatar to determine if avatar was uploaded by user or generated by google.
All default avatars contain '/mo/' string in their uri.

### DIFF
--- a/src/Two/GoogleProvider.php
+++ b/src/Two/GoogleProvider.php
@@ -85,6 +85,7 @@ class GoogleProvider extends AbstractProvider implements ProviderInterface
             'email' => Arr::get($user, 'email'),
             'avatar' => $avatarUrl,
             'avatar_original' => preg_replace('/\?sz=([0-9]+)/', '', $avatarUrl),
+            'avatar_is_default'  => strpos($avatarUrl, '/mo/') === false ? false : true,
         ]);
     }
 }


driesvints:
I'm not sure if it's even worth adding more Google Plus functionality as it's going to stop working as of March the 7th: https://developers.google.com/+/api-shutdown

No148:
> I'm not sure if it's even worth adding more Google Plus functionality as it's going to stop working as of March the 7th: https://developers.google.com/+/api-shutdown

It's not google+ feature as far as I understand. It's from google general account and gmail, I don't think it will go away. This parameter was available before google+ shutdown commits, so I thought it might be useful for everyone , because I really don't want to save default generated avatar and perform  a test against it in my application.

driesvints:
@No148 can you point to some api documentation for this?

No148:
@driesvints I can see that **gapi.auth2.BasicProfile** has **BasicProfile.getImageUrl()** in new documentstion. [https://developers.google.com/identity/sign-in/web/reference#googleauthcurrentuserget](url)
I guess they will keep avatars inside of google account in general. You can access it now by the way via [https://myaccount.google.com/personal-info](url) Now they are synchronized with google+ but I can't see any point to remove them entirely by google.

driesvints:
I can't see any hint anywhere at all that `/mo/` is an indicator for the default avatar. It's totally random afaik. Can you please give some proof or point to the exact spots in the docs where this is detailed?

taylorotwell:

PR #342, opened 2019-02-12
[2.0] Fix Google plus API breaking change

jackmcdade:
This is the same patch as submitted on the `3.x` and `4.x` branches, but now is compatible back to Laravel 5.1 for those that need it, because those people do indeed exist.

### DIFF
--- a/src/Two/GoogleProvider.php
+++ b/src/Two/GoogleProvider.php
@@ -58,7 +58,7 @@ class GoogleProvider extends AbstractProvider implements ProviderInterface
      */
     protected function getUserByToken($token)
     {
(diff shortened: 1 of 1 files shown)


driesvints:
Think we might need to drop hhvm testing as well. I'll do that in a followup commit.

driesvints:
Thanks!

jackmcdade:
Thanks @driesvints!

driesvints:
@jackmcdade fyi, the url seemed to be the wrong one and someone sent in a fix for it: https://github.com/laravel/socialite/pull/343

You might want to run a composer update for that.

jackmcdade:
Shoot, thanks both of you!

Issue #337, opened 2019-02-04
Make sure Google Provider is compatible with Google Sign-in


driesvints:
Since the old Google Plus functionality is going to be deprecated we'll need to make sure that we can use the suggested alternative Google Sign-in instead.

New OAuth2 functionality: https://developers.google.com/identity/protocols/OAuth2
Shutdown notice: https://developers.google.com/+/api-shutdown

No148:
@driesvints I tried to migrate to google sign-in using version 3 of socialite. I read the documentation and found this notes [https://developers.google.com/identity/sign-in/web/quick-migration-guide](url)
I have a custom socialtite provider so I was able to override methods from standart providers of socialite.
I created new api keys google sign-in, disabled google+ api in console and it worked as usually.

No148:
@eaverdeja I don't really know for sure at the moment... I've changed api key and created a specific api key for google sign-in. (which was only possible from this page https://developers.google.com/identity/sign-in/web/sign-in) After using it I just have my fingers crossed... But everything seems to be working fine right now. I wish I knew more, mate!

rajjanorkar:
- Just pull 3.0.x-dev
- No need to change anything in google developer console
- No need to change/create new api key/secrete etc.
- No other changes if you are not adding any additional scope (other than what package have)

It should just work fine.. i Tested it and it worked.

cincauhangus:
how can i make this compatible with laravel 5.2.x?

driesvints:
@cincauhangus 5.2.x isn't supported anymore. Please upgrade to the latest version.

Casperhr:
Any plans for a 2.x fix?

driesvints:
@Casperhr we actually pushed a 2.0 fix a couple of days ago. Use `2.0.x-dev` to pull it in.

driesvints:
Friendly reminder to not rely on `-dev` versions to long and to upgrade your apps eventually.

PR #346, opened 2019-02-22
[4.0] Update Google API urls

driesvints:
These updates ensure we're using the latest API urls from Google and continue to provide a working app after the shutdown of the Google+ functionality in March 2019.

By changing to the new v3 url for the user info retrieval, some of the keys from the response have changed. I've added the mappings to the old keys for the raw user data so nothing should break. These can be removed in v5.0 of Socialite.

Fixes https://github.com/laravel/socialite/issues/337

### DIFF
--- a/src/Two/GoogleProvider.php
+++ b/src/Two/GoogleProvider.php
@@ -37,7 +37,7 @@ class GoogleProvider extends AbstractProvider implements ProviderInterface
      */
     protected function getTokenUrl()
     {
-        return 'https://accounts.google.com/o/oauth2/token';
+        return 'https://www.googleapis.com/oauth2/v4/token';
     }
 
     /**
@@ -58,7 +58,7 @@ class GoogleProvider extends AbstractProvider implements ProviderInterface
      */
     protected function getUserByToken($token)
     {
-        $response = $this->getHttpClient()->get('https://www.googleapis.com/userinfo/v2/me?', [
+        $response = $this->getHttpClient()->get('https://www.googleapis.com/oauth2/v3/userinfo', [
             'query' => [
                 'prettyPrint' => 'false',
             ],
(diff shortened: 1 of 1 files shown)


stygiansabyss:
This is not present in 4.1.0 so google breaks again.

src/Two/GoogleProvider.php
```diff
@@ -76,15 +76,18 @@ protected function getUserByToken($token)
...
         return (new User)->setRaw($user)->map([
-            'id' => $user['id'],
+            'id' => Arr::get($user, 'sub'),
             'nickname' => Arr::get($user, 'nickname'),
             'name' => Arr::get($user, 'name'),
             'email' => Arr::get($user, 'email'),
-            'avatar' => $avatarUrl,
-            'avatar_original' => preg_replace('/\?sz=([0-9]+)/', '', $avatarUrl),
+            'avatar' => $avatarUrl = Arr::get($user, 'picture'),
+            'original_avatar' => $avatarUrl,
```
goyote: This is a backwards compat break. Changing the key avatar_original to original_avatar

goyote:
This is a backwards compat break. Changing the key avatar_original to original_avatar

src/Two/GoogleProvider.php
bkuhl: Unfortunately Laravel packages do not follow semantic versioning so things can break at any time.

goyote:
But what's the point of changing the array key name for no reason?

src/Two/GoogleProvider.php
driesvints: All packages by Laravel except Lumen and the framework itself follow SemVer. This was a mistake. I'll correct it.

driesvints:
@bkuhl @goyote released a bug fix for this.

bkuhl:
Oh great, thanks!

PR #350, opened 2019-03-02
Update request to pass token as header for Github Provider

WyattCast44:
Update user and email API URLS to pass the access token as a header, as suggested in the GitHub documentation.

Source: https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/


taylorotwell:
If nothing is broken with the current implementation I think I will leave it as-is to avoid breaking changes. /cc @driesvints

driesvints:
The actual phrase referenced is:

> You can pass the token in the query params as shown above, but a cleaner approach is to include it in the Authorization header.

So I'd just leave things be since there's nothing broken here.

Issue #369, opened 2019-06-09
Sign in with apple id


kevincobain2000:
Since the announcement of sign in with apple id. It'd be good to have it as a socialite provider for sign in with apple.

https://developer.apple.com/sign-in-with-apple/

olivernybroe:
> We are not accepting new adapters.
> Adapters for other platforms are listed at the community driven Socialite Providers website.

So you should prob. create the issue in that package instead https://github.com/SocialiteProviders/Providers or just create your own package with it.

driesvints:
@olivernybroe I realize that this is in the readme but I suspect that this will be such a widely adopted adapter so we could maybe consider adding it in. 

Might need https://developer.apple.com/documentation/signinwithapplejs ?

taylorotwell:
Feel free to PR.

m1guelpf:
[This PR to SocialiteProviders/Providers](https://github.com/SocialiteProviders/Providers/pull/315) might be a good starting point.

theodh:
The email address will be available soon, see issue:
https://forums.developer.apple.com/thread/118209

We need to wait after the update from Apple, but we could implement this @jbrooksuk #371 (comment)
with a dummy param email_id and change the variable name after the update, so we could be live quickly.

Shall I make a PR from your repo @pxgamer https://github.com/pxgamer/socialite/tree/feature/apple-provider
so we could go live after the update form apple? Or do you want to implement this?

Perhaps a suggestion for the socialite documentation from apple with there anonymous-email-feature:

>In order to contact users that use Apple’s private email relay service, you need to register domains and email addresses that your organization will use for communication. Domains and domains associated with email addresses must comply with Sender Policy Framework standards and be verified by Apple before they are successfully registered.

Update From https://github.com/laravel/socialite/pull/371 message @pxgamer, will start a PR from his repo and snippet @jbrooksuk

theodh:
The snippet has been implemented see https://github.com/theodh/socialite/commit/bc70f9a06e307bb7dfa0a52987138b942fe0c469
 At this moment the scope is only name. The state at this moment: a unique verifier user will be created. If we add email in the scope it will throw an exception until Apple has added this param. If email is a requirement for socialite we should wait after the update, otherwise I could do a PR. Other suggestions?

pxgamer:
[...]

PR #381, opened 2019-07-17
[4.0] Apple issue 369

theodh:
#369 The apple provider has some different behaviour as a 'common' oauth provider, perhaps notes in the documentation:
* You only get the email address in the first login of the user. You should save the email address (user->email) and the apple identifier (user->id). The second time you use this identifier to find the user in your laravel applications. 
* Clientsecret will expire, the user should maintain this.
* The apple user could use the anonymous email feature. To sent emails please configure
https://developer.apple.com/account/resources/services/configure
> In order to contact users that use Apple’s private email relay service, you need to register domains and email addresses that your organization will use for communication. Domains and domains associated with email addresses must comply with Sender Policy Framework standards and be verified by Apple before they are successfully registered.


driesvints:
Isn’t it better that we delay this feature until the final ios versions have been released? Not sure how much will change until then?

theodh:
@driesvints 
> Isn’t it better that we delay this feature until the final ios versions have been released? Not sure how much will change until then?

Good point, other view, we should test this in some environments. I'm already having this provider in the production environment. I'm also monitoring the log. My suggestion a beta-branch in your repo, or that other developer could use a VCS to mine. 

Laravel app: composer.json

Change require:
```
"laravel/socialite": "dev-apple-issue-369",
```
Add
```
"repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/theodh/socialite"
        }
    ]
```

I need some testers :guardsman: :trumpet:

driesvints:
@theodh it's best that you add this to the related issue so we can keep track of all the discussion in one single place.

PR #386, opened 2019-08-14
Documenting @magic static method

danieldevsquad:
Documenting the magic static `driver` method in `Laravel\Socialite\Facade\Socialite` class.

```php
@method static Provider driver(string $driver = null)
```

A little bit of IntelliSense support should not hurt anyone ;)


taylorotwell:
We always use fully qualified class names in doc blocks.

PR #390, opened 2019-08-23
[4.0] Prepare for Laravel 6.0 release

driesvints:


GrahamCampbell:
Is it not wrong to allow installation on Laravel 7, when it may not actually be compatible?

driesvints:
@grahamcampbell this is the only way for apps who use socialite and want to test v7.0 already to do that. We always make sure the builds for next releases are passing (as I did today) before doing a major release

GrahamCampbell:
> this is the only way for apps who use socialite and want to test v7.0 already to do that. 

It's not the only way. You can just require Laravel 7 and tell composer you want to pretend it's Laravel 6 using the "as" notation. That happens in your app composer.json, which is much better than having incorrect version constraits out there in wild. You might as well just put `*` for the version, otherwise. :P

driesvints:
@GrahamCampbell Taylor prefers it this way.

GrahamCampbell:
:/

GrahamCampbell:
Defeats the object of vesion constraints.

PR #408, opened 2019-11-15
[4.x] Change default graph version, manually set desired version

jdrzejb:
It changes default graph version to 3.3.

Also, it introduces a way of changing the default version. I'm still having second thoughts regarding this, as it might break functionality if facebook decides to change how object is accessed, and we will need a proper mapping between graph versions. 

The usage I have in mind:

```php
return Socialite::driver('facebook')->usingGraphVersion('v5.0')->redirect();
```

Let me know what do you think

### DIFF
--- a/src/Two/FacebookProvider.php
+++ b/src/Two/FacebookProvider.php
@@ -178,11 +178,11 @@ class FacebookProvider extends AbstractProvider implements ProviderInterface
 
 
     /**
-     * Replace default graph version to use with given request
+     * Replace default graph version to use with given request.
      *
      * @return $this
      */
-    public function usingGraphVersion(string $version) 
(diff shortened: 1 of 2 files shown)


driesvints:
Can you guarantee that updating the default API to 3.3 from 3.0 doesn't breaks anything?

jdrzejb:
There are no breaking changes in methods that are used in the package between `v3.0` and `v3.3`.
I did not use it though. If you wish, I can test it live today when preparing social auth in my app.

https://developers.facebook.com/docs/graph-api/changelog/version3.3
https://developers.facebook.com/docs/graph-api/changelog/version3.2
https://developers.facebook.com/docs/graph-api/changelog/version3.1

PR #411, opened 2019-11-26
[4.x] Allow route/action based redirect url definition

netpok:
Currently one can only define the redirect route by entering it to the config file because the router is not available when the config files are loaded (or by passing a callback but that makes the config non-serializable).

This pull request makes possible to use either routes or actions by passing them in an array form:
```php
['route' => 'name']; // named route
['action' => 'controller@method']; // route based on action
['route' => 'name', 'parameters' => ['random', 'stuff']]; // with additional parameters
```

As an alternative I have considered introducing a redirect_route key but I think this is a superior solution.

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -167,6 +167,20 @@ class SocialiteManager extends Manager implements Contracts\Factory
     {
         $redirect = value($config['redirect']);
 
+        if (is_array($redirect)) {
+            if (array_key_exists('route', $redirect)) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
I am closing this pull request because it lacks sufficient explanation, **tests**, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Feel free to re-submit your change with a **thorough** explanation of the feature and tests - integration tests are preferred over unit tests. Please include it's benefit to end users; the reasons it does not break any existing features; how it makes building web applications easier, etc.

Thanks!

PR #418, opened 2020-01-06
3.0

mertcagiran:


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "laravel/socialite",
+    "name": "tevfikmert/socialite",
```
GrahamCampbell: Please don't change the name of the package. Note also, that it not recommended practice to publish forks on packagist.

GrahamCampbell: Please don't change the name of the package. Note also, that it not recommended practice to publish forks on packagist.

PR #420, opened 2020-01-25
[4.x] Replace ArrayAccess interface calls with Container::make

netpok:
Since the SocialiteManager class accepts an instance of the Container interface which not implements ArrayAccess it should use the `make` method instead of array access.

This pull request updates the code to use the `make` method where necessary.

Actually it probably shouldn't use ArrayAccess on the config instance either, but it is not defined anywhere that the make('config') can return an object that only implements the Repository interface and not the base class itself.


taylorotwell:
Nothing is broken here.

Issue #388, opened 2019-08-19
Use route for redirect URL


yakatz:
I have general routes to allow easily adding Socialite providers:
```php
Route::get('login/{provider}', 'Auth\LoginController@redirectToProvider')->name('socialite.redirect');  
Route::get('login/{provider}/callback', 'Auth\LoginController@handleProviderCallback')->name('socialite.callback');
```

Because config is loaded before routes, it isn't possible to set the `redirect` option in `services.php` using something like `route('socialite.callback', ['provider' => 'google']),`.

My proposal is to allow setting a route name for Socialite and use that to create a default `redirect` if one isn't specified.

Something like this (although there isn't currently a simple way to get the provider name either?):
```php
    /**
     * Format the callback URL, resolving a relative URI if needed.
     *
     * @param  array  $config
     * @return string
     */
    protected function formatRedirectUrl(array $config)
    {
        $redirect = value($config['redirect']);
        if ( ! defined($redirect) ) {
            return $this->defaultRedirectUrl($config);
        }
        return Str::startsWith($redirect, '/')
                    ? $this->app['url']->to($redirect)
                    : $redirect;
    }

    /**
     * Get a default callback route.
     *
     * @param  array  $config
     * @return string
     */
    protected function defaultRedirectUrl(array $config)
    {
        $socialite_config = $this->app['config']['socialite'];
        $route = value($config['route']);
        if ($route) {
            return $this->app['url']->route($route, ['provider' => $config['provider'] ]);
        } else {
            throw new Exception();
        }
    }
```

driesvints:
So this is basically an easier way instead of setting up the controller for routing? https://laravel.com/docs/5.8/socialite#routing

yakatz:
No, this is substantially different. Right now you need to set the callback url in `serices.php` individually for each provider and there is no way to set it in the controller. Since you need to have a route for the callback anyway, this change would allow you to not need to specify the callback in the configuration.

driesvints:
Feel free to send in a PR when you have some time.

vijayq9:
Open vendor\laravel\socialite\src\SocialiteManager.php and  Replace
  protected function createFacebookDriver()
    {
         $config = $this->app['config']['services.facebook'];
          return $this->buildProvider(
            FacebookProvider::class, $config
        );
    }

      protected function createFacebookDriver()
    {
         $config = $this->app['config']['services.stripe.facebook'];
          return $this->buildProvider(
            FacebookProvider::class, $config
        );
[...]

PR #421, opened 2020-01-25
[4.x] Allow route/action based redirect url definition

netpok:
Currently one can only define the redirect route by entering it to the config file because the router is not available when the config files are loaded (or by passing a callback but that makes the config non-serializable). This means that the urls have to be defined at least two places and updated later if changed.

This pull request makes possible to use either routes or actions by passing them in an array form:

```php
['route' => 'name'];
```
Generates an URL for a named route.

```php
['action' => 'controller@method'];
```
Generates an URL for a controller action.

```php
['route' => 'name', 'parameters' => ['random', 'stuff']]);
```
Additional parameter can be also supplied just like to the route method.

As an alternative I have considered introducing a redirect_route key but I think this is a superior solution.

Follow up to #411, solves #388

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -35,7 +35,7 @@ class SocialiteManager extends Manager implements Contracts\Factory
      */
     protected function createGithubDriver()
     {
-        $config = $this->app['config']['services.github'];
+        $config = $this->app->make('config')['services.github'];
 
         return $this->buildProvider(
             GithubProvider::class, $config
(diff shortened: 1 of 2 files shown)


taylorotwell:

yakatz:
@taylorotwell I am not sure how it would be possible to release this as a package since it modifies the redirectURL set in the query string for these providers. This PR helps keep configuration DRY because otherwise a separate redirect URL must be set for every provider in the configuration. Please reconsider.

Issue #427, opened 2020-02-04
Update Github authentication


driesvints:
Got this email from Github:

> Hello there!
> 
> On February 4th, 2020 at 00:53 (UTC) your application (Laravel.io) used an access token (with the User-Agent GuzzleHttp/6.3.3 curl/7.47.0 PHP/7.3.10-1+ubuntu16.04.1+deb.sury.org+1) as part of a query parameter to access an endpoint through the GitHub API.
> 
> https://api.github.com/user
> 
> Please use the Authorization HTTP header instead as using the `access_token` query parameter is deprecated.
> 
> Depending on your API usage, we'll be sending you this email reminder once every 3 days for each token and User-Agent used in API calls made on your behalf.
> Just one URL that was accessed with a token and User-Agent combination will be listed in the email reminder, not all.
> 
> Visit https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters for more information.
> 
> Thanks,
> The GitHub Team

We'll need to update the Github provider for this.

yemilgr:
I got the same email too just hours ago...

PR #428, opened 2020-02-04
use Authorization header instead of using the `access_token` query parameter

WessamSaeid:
Github is deprecating using `access_token` as query parameter and they recommend sending OAuth tokens using the Authorization header.
as mentioned in this issue #427 

see https://developer.github.com/v3/#oauth2-token-sent-in-a-header


driesvints:
Hey @WessamSaeid, can you send this to the 4.0 branch instead? Thanks!

PR #430, opened 2020-02-04
[4.x] use Authorization header instead of using the `access_token` query parameter

WessamSaeid:
Github is deprecating using access_token as query parameter and they recommend sending OAuth tokens using the Authorization header.
as mentioned in this issue #427

see https://developer.github.com/v3/#oauth2-token-sent-in-a-header


robbieaverill:
Thanks all for the prompt action/merge/release 👍

PR #431, opened 2020-02-12
fix: use Authorization header instead of using the `access_token`

fmendozaro:
Github is deprecating using access_token as a query parameter and they recommend sending OAuth tokens using the Authorization header for version 2.X. The patch was applied only to the latest version.

Based on this PR: #430


GrahamCampbell:
Socialite 2.0 is not maintained. Please upgrade to 4.0 or higher.

PR #432, opened 2020-02-17
[3.x] Use Authorization header instead of using the `access_token` query parameter

DCzajkowski:
Mirror of #430, but for the `3.0` branch. I understand 3.x may not be the newest, but 4.x requires Laravel upgrade and that is not really viable in our case. Please consider releasing this, as it is a very small change. Thank you.


driesvints:
Socialite 3 isn't supported anymore.

PR #438, opened 2020-03-18
Replaced deprecated app property in SocialiteManager

ctwillie:
…Manger class

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -35,7 +35,7 @@ class SocialiteManager extends Manager implements Contracts\Factory
      */
     protected function createGithubDriver()
     {
-        $config = $this->app['config']['services.github'];
+        $config = $this->container['config']['services.github'];
 
         return $this->buildProvider(
             GithubProvider::class, $config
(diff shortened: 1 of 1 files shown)


driesvints:
We'll need to convert the array access calls to `make` method calls because the container doesn't implements `ArrayAccess`.

ctwillie:
I believe it does. According to the api docs.

https://laravel.com/api/7.x/Illuminate/Container/Container.html

ctwillie:
@driesvints 

also try this

```php

$relectionContainer = new ReflectionClass(get_class(app()));

dd(relectionContainer->getInterfaceNames());
```
And you will see 'ArrayAccess' in the list

driesvints:
@ctwillie that's the implementation. The property references the contract which doesn't implements it.

ctwillie:
@driesvints Aha! Something I'll keep in mind from here on out. Thanks. Will do a PR.

GrahamCampbell:
@ctwillie those are not the docs for the container contract though. ;)

https://laravel.com/api/7.x/Illuminate/Contracts/Container/Container.html

ctwillie:
@GrahamCampbell  yea, I realize that now. I learned something new to pay attention to and stand corrected. Give me a break, I've only been coding for a couple of years, lol. Thanks for pointing that out though.

driesvints:
@ctwillie we had to revert this because it broke apps running on Laravel 5.7 & 5.8. We can do these changes on master where only Laravel 6 and 7 are supported.

https://github.com/laravel/socialite/pull/450

ctwillie:
Ok cool. Will do the same in master branch.

Issue #442, opened 2020-04-22
Multiple concurrent OAuth2 callback requests can get valid state parameter value


lebedevsergey:
- Socialite Version: 3.4.0
- Laravel Version: 5.8
- PHP Version: 7.2
- Database Driver & Version:

### Description:
Multiple concurrent OAuth2 callback requests can get valid state parameter value which stores in `session` and is removed (pulled) immediately after reading it. This seems to be related to the way that `Laravel` work with session data - all changes in session data are saved to the permanent store only [at the end of request handling](github://https.com/laravel/framework/issues/30996
), so a the other coming concurrent request can get the pulled value.  

### Steps To Reproduce:
Create a typical OAuth2 workflow with callback which uses `state` parameter for request validation. OAuth2 authorization service is not important, hovewer we faced and reproduced this issue with Mixer OAuth2. When OAuth2 callback gets multiple concurrent requests at the same time (for example when user quickly pushe multiple times "Approve application permissions" button in [Mixer](http://mixer.com)), valid `state` parameter value gets more that one callback request. 
This can be the cause of different issues in service that uses OAuth2 authorization. For example, in our service this results in registration of multiple users from the same [Mixer](http://mixer.com) account.

driesvints:
Heya, thanks for reporting. I'm currently running a fairly large app with thousands of users and never experienced this problem. This seems like an edge case really. I guess we can't prevent people from pressing a button multiple times? I don't see an easy way to solve this in Socialite but you're free to attempt a new PR.

wb-joona:
Laravel supports [Session Blocking](https://laravel.com/docs/9.x/session#session-blocking) nowadays (since 7.x maybe?), which lets you limit concurrent requests to a specific route.

lebedevsergey:
Thank you @wb-joona, we'll consider using this feature to prevent our issues

PR #443, opened 2020-04-22
Fix issue "Multiple concurrent OAuth2 callback requests can get valid state parameter value"

lebedevsergey:
Fix [this](https://github.com/laravel/socialite/issues/442) issue by saving `session` data immediately after `state` parameter pulling


taylorotwell:
Honestly that is going to cause other problems. The `save` method ages flash data and changes some state in the Session Store (`started = false`).

lebedevsergey:
@taylorotwell, thank you for the answer, is there a way to flush current session state without these issues?

PR #449, opened 2020-05-29
Support Guzzle 7

GrahamCampbell:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,13 +17,13 @@
     "require": {
         "php": "^7.1.3",
         "ext-json": "*",
-        "guzzlehttp/guzzle": "~6.0",
-        "illuminate/http": "~5.7.0|~5.8.0|^6.0|^7.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -17,13 +17,13 @@
...
-        "illuminate/http": "~5.7.0|~5.8.0|^6.0|^7.0",
-        "illuminate/support": "~5.7.0|~5.8.0|^6.0|^7.0",
-        "league/oauth1-client": "~1.0"
+        "guzzlehttp/guzzle": "^6.0|^7.0",
+        "illuminate/http": "^5.7|^6.0|^7.0",
+        "illuminate/support": "^5.7|^6.0|^7.0",
+        "league/oauth1-client": "^1.0"
     },
     "require-dev": {
-        "illuminate/contracts": "~5.7.0|~5.8.0|^6.0|^7.0",
```
driesvints, line 26: I realise ^ will allow for both 5.7 and 5.8 but for the sake of consistency can we please leave these constraints as is?

PR #465, opened 2020-09-03
[5.x] Support Laravel 8

driesvints:


composer.json
```diff
@@ -17,15 +17,15 @@
     "require": {
         "php": "^7.2",
```
sergiy-petrov, line 18: @driesvints shouldn't this be changed to "^7.2|^8.0" ?

sergiy-petrov: just like here
https://github.com/laravel/horizon/commit/2d55a1fb776bf945f46b8b31efac5d7611afae37

driesvints: We haven't updated socialite for PHP 8 yet. This PR is 2 months old.

PR #467, opened 2020-09-04
Fix guzzle version

ankurk91:
Hi,

There is no guzzle v8.0 yet 😄


GrahamCampbell:
// cc @driesvints

driesvints:
Thanks

Issue #479, opened 2020-10-01
Facebook 403 Forbidden Missing Permissions


raditzfarhan:
- Socialite Version: v5.0.1
- Laravel Version: 8.7.1
- PHP Version: 7.4.6
- Database Driver & Version: MySQL 5.7.24

### Description:
When using Facebook driver to login, received 403 Forbidden error on callback because of the following code below.

`Socialite::driver('facebook')->user()`

Exact error message as below:
```
{
  "error": {
    "message": "(#200) Missing Permissions",
    "type": "OAuthException",
    "code": 200,
    "fbtrace_id": "AnDmOjgOB6O56X0cpM0y-rO"
  }
}
```

### Steps To Reproduce:
1. Install latest Laravel.
2. Install latest Laravel Socialite plugin.
3. Setup your redirect and callback routes and configure your Facebook app credentials.
4. Set driver as `facebook`.
5. Access the login/redirect URL to authenticate with Facebook.
6. On your callback, call `Socialite::driver('facebook')->user()`

mathiasmoser:
we get the same error with Laravel 7.28.1 and socialite v.4.4.1

mathiasmoser:
it seems that there is a permission problem with the requested fields (fields=name,email,gender,verified,link), if i manually remove gender and link from the fields i get a response from fb

mathiasmoser:
as a workaround i removed the fields gender and link from the requested fields, because this permissions have be requested in the facebook app settings and also require separate review process:

**With these reduced fields the request will work again:**
Socialite::driver('facebook')->fields(['name', 'email', 'verified'])->userFromToken($accessToken)

driesvints:
I want to investigate first why this is broken in the first place after having worked for so many years. Did Facebook update its API? The API is versioned so this would be a breaking change on their part? Seems unlikely.

raditzfarhan:
@driesvints I believe so. Yesterday was still working fine.

raditzfarhan:
> as a workaround i removed the fields gender and link from the requested fields, because this permissions have be requested in the facebook app settings and also require separate review process:
> 
> **With these reduced fields the request will work again:**
> Socialite::driver('facebook')->fields(['name', 'email', 'verified'])->userFromToken($accessToken)

Thanks @mathiasmoser your solution is working.

Maxwellkurt0407:
> 'name', 'email', 'verified'

Good day, how do I apply this fix? I've been having the same problem as well

[...]

PR #481, opened 2020-10-01
Fix Facebook 403 Forbidden Missing Permissions

raditzfarhan:
Fix Facebook 403 Forbidden Missing Permissions because of `link` field. Related to issue #479


driesvints:
Let's first investigate what broke it.

PR #486, opened 2020-10-12
Create method inCallbackUri()

jotaelesalinas:
Create the method `inCallbackUri()` in order to be able to add parameters to the `callback_uri` parameter of the redirection URL.

Similar to `with()`, but instead of appending the paramenters to the whole redirection URL, they are appended to the `callback_uri` parameter.

As an example, imagine that we have the following data from the provider example.com:

```
CLIENT_ID=1234567890
CALLBACK_URI=http://myweb.com/oauth-callback
```

Generated redirection URL with standard `$socialite_provider->redirect()`:

```
https://example.com/oauth/authorize
    ?client_id=123567890
    &redirect_uri=http%3A%2F%2Fmyweb.com%2Foauth-callback
```

Now we want to send a key/value pair and get it back.

A simple use case could be indicating the route where the authorisation was started, in order to redirect to the proper route afterwards: `from=login`, `from=profile` or `from=manage-auths`...

Generated redirection URL with **existing method `with()`**:

```php
$socialite_provider->with(['from' => 'login'])->redirect()`:
```

```
https://example.com/oauth/authorize
    ?client_id=123567890
    &redirect_uri=http%3A%2F%2Fmyweb.com%2Foauth-callback
    &from=login
```

Here, `from=login` is passed to the provider, but is lost in the response.

Generated redirection URL with **new method `inCallbackUri()`**:

```php
$socialite_provider->inCallbackUri(['from' => 'login'])->redirect();
```

```
https://example.com/oauth/authorize
    ?client_id=123567890
    &redirect_uri=http%3A%2F%2Fmyweb.com%2Foauth-callback%3Ffrom%3Dlogin
```

Here, `from=login` is passed to the provider and sent back to us in the response!

The callback URL (parameter `redirect_uri` in the redirection URL), once urldecoded, is:

```
http://myweb.com/oauth-callback?from=login
```

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -54,6 +54,15 @@ abstract class AbstractProvider implements ProviderContract
     protected $parameters = [];
 
     /**
+     * The custom parameters to be sent inside the callback URI parameter of the request.
+     * This way, parameters can be passed two-way from the function starting the redirection
+     * to the provider and back to the function handling the response.
+     *
+     * @var array
+     */
+    protected $callbackUriParameters = [];
+
+    /**
      * The scopes being requested.
      *
      * @var array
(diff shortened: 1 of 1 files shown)


taylorotwell:
You could just put a variable in the session instead to preserve a value for yourself.

jotaelesalinas:
That’s what I am doing now, but I just don’t like it. Thanks for all this, anyway!

PR #491, opened 2020-10-26
[5.x] Added merge of parameters to getTokenFields

kevincam:
This allows token fields to be overridden at runtime for methods like user() the same way as the initial redirect(). Currently redirect_uri will throw an error when fetching user data with: Socialite::driver($driver)->with(['redirect_uri' => $redirect])->user()

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -273,13 +273,15 @@ abstract class AbstractProvider implements ProviderContract
      */
     protected function getTokenFields($code)
     {
-        return [
+        $fields = [
             'grant_type' => 'authorization_code',
             'client_id' => $this->clientId,
             'client_secret' => $this->clientSecret,
             'code' => $code,
             'redirect_uri' => $this->redirectUrl,
         ];
+
+        return array_merge($fields, $this->parameters);
     }
 
     /**


taylorotwell:
Why would you need to do this? More explanation?

kevincam:
I'm doubling up the use of the facebook oauth for facebook login and also instagram stats, since instagram's API now no longer lets you fetch any useful data. Because I am handling the callback differently, I am passing a custom redirect at runtime based on which source you are connecting:

Socialite::driver('facebook')->with(['redirect_uri' => $redirect])->redirect();

This works perfectly, but then there's no way to do it for the user() call:

Socialite::driver('facebook')->with(['redirect_uri' => $redirect])->user();

If the redirect URI mismatches then facebook will reject the token.

This change will let you pass a separate URI (or other custom data) at runtime to the user method the same way you can do it for the redirect, so it will stay consistent if the API expects it.

driesvints:
The `with` was never meant to override reserved keywords like this. See the note at https://laravel.com/docs/8.x/socialite#optional-parameters. You can already set this value through your config and with the `redirectUrl` method.

Besides this is a breaking change for anyone already using the `with` method. I don't think this should be merged, sorry.

kevincam:
Though it is not meant to override reserved keywords, it works quite well for passing this sort of runtime change to the redirect URI. It just doesn’t work entirely because it is missing on one method. Currently it seems inconsistent that `with` is not usable with `user` even just for passing other arbitrary  custom values to that call.

taylorotwell:
Will close for now pending feedback from Dries.

PR #508, opened 2020-12-24
[5.x] Fix Gitlab provider auth url with scopes

### LINKED ISSUE #507: Gitlab scopes not working
- Socialite Version: 5.1.2
- Laravel Version: 8.19.0
- PHP Version: 8.0.0
- Database Driver & Version: MySQL 8.0

### Description:

Defining scopes with Gitlab provider not working because of an incorrect URL parameter format.

The Gitlab scope separator should be `+` and not `,` like it is by default.

But there is a second problem. The `+` character is transformed into `%2B` when the URL is created by `http_build_query`.

### Steps To Reproduce:

Specify two scopes in the redirect action:

```php
return Socialite::driver('gitlab')
    ->setScopes(['read_user', 'read_api'])
    ->redirect();
```

Result this error:

image

Because of this parameter `&scope=read_user%2Bread_api` in the URL.
Should be `&scope=read_user+read_api`.

jdecool:
This PR is a proposal to fix #507 

When defining scopes with Gitlab provider not working because of an incorrect URL parameter format.

The Gitlab scope separator should be `+` and not `,` like it is by default.

But there is a second problem. The `+` character is transformed into `%2B` when the URL is created by `http_build_query`.

### DIFF
--- a/src/Two/GitlabProvider.php
+++ b/src/Two/GitlabProvider.php
@@ -12,11 +12,27 @@ class GitlabProvider extends AbstractProvider implements ProviderInterface
     protected $scopes = ['read_user'];
 
     /**
+     * The separating character for the requested scopes.
+     *
+     * @var string
+     */
+    protected $scopeSeparator = '+';
+
+    /**
      * {@inheritdoc}
      */
     protected function getAuthUrl($state)
     {
-        return $this->buildAuthUrlFromBase('https://gitlab.com/oauth/authorize', $state);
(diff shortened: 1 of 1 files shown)


driesvints:
Are you 100% sure this won't break older GitLab OAuth2 apps? It seems to me something changed on GitLab's part overnight and this change could break older apps.

driesvints:
Closing as the original issue isn't verified yet.

Issue #483, opened 2020-10-09
Add support for the OAuth 2.0 PKCE extension


aaronpk:
Many OAuth providers support the [PKCE](https://oauth.net/2/pkce/) extension, so it would be great to see this added in the core OAuth 2 class so that it can be leveraged by providers.

While PKCE was originally created for native apps, it also protects against [authorization code injection](https://tools.ietf.org/html/draft-ietf-oauth-security-topics-16#section-4.5) even for web server apps that use a client secret.

How to implement:
* Generate a new random string and store it in the session similar to how Socialite currently [stores the `state`](https://github.com/laravel/socialite/blob/5.x/src/Two/AbstractProvider.php#L150) value. 
* Then generate a SHA256 hash of that string and include that in the [`getCodeFields`](https://github.com/laravel/socialite/blob/5.x/src/Two/AbstractProvider.php#L175) parameters. 
* Include the original random string in the [`getTokenFields`](https://github.com/laravel/socialite/blob/5.x/src/Two/AbstractProvider.php#L274) POST request.

driesvints:
https://laravel.com/docs/8.x/passport#code-grant-pkce

aaronpk:
That's a link to Passport. I'm talking about adding support as an OAuth client so that Socialite providers can use PKCE without having to write that code in each provider's extension.

driesvints:
Ah wow sorry. Totally didn't see this was the Socialite repo. 

I'm not sure if that's needed though. Socialite is only for social authentication and isn't meant to authorize API's.

aaronpk:
No worries.

PKCE isn't just about authorizing APIs, it's for any OAuth client regardless of what it's doing. Since Socialite is used to talk to many OAuth servers (even just for social authentication), the client should be doing PKCE wherever possible. I thought it would make the most sense to add this at the core rather than duplicate that work in every Socialite provider.

driesvints:
Do all of the 1st party providers currently support this? What key does it need to be stored under?

aaronpk:
Unfortunately the only one of the providers in this repo that supports it is Google. However many of the ones in the [SocialiteProviders](https://github.com/SocialiteProviders/Providers/tree/master/src) repo support it.

That said, any conformant OAuth server is supposed to not fail if the additional PKCE parameters are included (PKCE was designed to be backwards compatible), although I have seen at least one server error out anyway, so I wouldn't recommend always enabling PKCE.

The parameters are called `code_verifier` and `code_challenge`.

[...]

PR #518, opened 2021-02-11
[5.x] Add support for OAuth 2.0 PKCE extension

aaronpk:
This PR adds support for the OAuth 2.0 [PKCE](https://oauth.net/2/pkce/) extension, as discussed in #483. This extension is recommended in the latest version of the [OAuth 2.0 Security Best Current Practice](https://oauth.net/2/oauth-best-practice/). 

* Adds the PKCE logic in the core OAuth 2 provider
* Enables PKCE with Google
* Other providers can enable PKCE by setting the new `usesPKCE` property on the provider
* The default value is `false`, which is the safer way to handle backwards compatibility, even though technically OAuth providers are supposed to ignore unknown parameters so it should be theoretically possible to always send the PKCE parameters anyway

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -475,7 +475,8 @@ abstract class AbstractProvider implements ProviderContract
     {
         $verifier = $this->request->session()->pull('code_verifier');
         $hashed = hash('sha256', $verifier, true);
-        return rtrim(strtr(base64_encode($hashed), '+/', '-_'), '='); // base64-URL-encoding
+
+        return rtrim(strtr(base64_encode($hashed), '+/', '-_'), '=');
     }
 
     protected function getCodeChallengeMethod() 
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -82,6 +82,13 @@ abstract class AbstractProvider implements ProviderContract
     protected $stateless = false;
 
     /**
+     * Indicates if PKCE should be used.
+     *
+     * @var bool
+     */
+    protected $usesPKCE = false;
+
+    /**
      * The custom Guzzle configuration options.
      *
      * @var array
@@ -158,6 +165,10 @@ abstract class AbstractProvider implements ProviderContract
             $this->request->session()->put('state', $state = $this->getState());
         }
 
+        if ($this->usesPKCE()) {
+            $this->request->session()->put('code_verifier', $codeVerifier = $this->getCodeVerifier());
+        }
+
         return new RedirectResponse($this->getAuthUrl($state));
     }
 
@@ -190,6 +201,11 @@ abstract class AbstractProvider implements ProviderContract
 
         if ($this->usesState()) {
             $fields['state'] = $state;
+        }
+
+        if ($this->usesPKCE()) {
+            $fields['code_challenge'] = $this->getCodeChallenge();
+            $fields['code_challenge_method'] = $this->getCodeChallengeMethod();
         }
 
         return array_merge($fields, $this->parameters);
@@ -284,13 +300,19 @@ abstract class AbstractProvider implements ProviderContract
      */
     protected function getTokenFields($code)
     {
-        return [
+        $fields = [
             'grant_type' => 'authorization_code',
             'client_id' => $this->clientId,
             'client_secret' => $this->clientSecret,
             'code' => $code,
             'redirect_uri' => $this->redirectUrl,
         ];
+
+        if ($this->usesPKCE()) {
+            $field['code_verifier'] = $this->request->session()->pull('code_verifier');
+        }
+
+        return $fields;
     }
 
     /**
@@ -434,6 +456,32 @@ abstract class AbstractProvider implements ProviderContract
         return Str::random(40);
     }
 
+    /**
+     * Determine if the provider uses PKCE.
+     *
+     * @return bool
+     */
+    protected function usesPKCE()
+    {
+        return $this->usesPKCE;
+    }
+
+    protected function getCodeVerifier()
+    {
+        return Str::random(96);
+    }
+
+    protected function getCodeChallenge() 
+    {
+        $verifier = $this->request->session()->pull('code_verifier');
+        $hashed = hash('sha256', $verifier, true);
+        return rtrim(strtr(base64_encode($hashed), '+/', '-_'), '='); // base64-URL-encoding
+    }
+
+    protected function getCodeChallengeMethod() 
+    {
+        return 'S256';
+    }
     /**
      * Set the custom parameters of the request.
      *
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -466,12 +466,22 @@ abstract class AbstractProvider implements ProviderContract
         return $this->usesPKCE;
     }
 
+    /**
+     * Generates a random string of the right length for the PKCE code verifier.
+     *
+     * @return string
+     */
     protected function getCodeVerifier()
     {
         return Str::random(96);
     }
 
-    protected function getCodeChallenge() 
+    /**
+     * Generates the PKCE code challenge based on the PKCE code verifier in the session.
+     *
+     * @return string
+     */
+    protected function getCodeChallenge()
     {
         $verifier = $this->request->session()->pull('code_verifier');
         $hashed = hash('sha256', $verifier, true);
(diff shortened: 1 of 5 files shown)


tests/OAuthTwoTest.php
```diff
@@ -24,17 +25,67 @@ protected function tearDown(): void
...
         $provider = new OAuthTwoTestProviderStub($request, 'client_id', 'client_secret', 'redirect');
         $response = $provider->redirect();
 
         $this->assertInstanceOf(SymfonyRedirectResponse::class, $response);
         $this->assertInstanceOf(RedirectResponse::class, $response);
-        $this->assertSame('http://auth.url', $response->getTargetUrl());
+        $this->assertSame('http://auth.url?client_id=client_id&redirect_uri=redirect&scope=&response_type=code&state='.$state, $response->getTargetUrl());
+    }
+
+    private static $codeVerifier = null;
```
aaronpk, line 52: I was unclear on how to get this value stored in the test in a local variable, it seemed to only work when stored as a static class variable. The test needs to know the value the main library generated, and also needs to be able to return it to the main library when pulled from the session. Let me know if there is a better way to do this, it was very difficult to find any docs on how to do this kind of test.

src/Two/AbstractProvider.php
```diff
@@ -434,6 +456,32 @@ protected function getState()
...
+    }
+
+    protected function getCodeChallenge() 
+    {
+        $verifier = $this->request->session()->pull('code_verifier');
+        $hashed = hash('sha256', $verifier, true);
+        return rtrim(strtr(base64_encode($hashed), '+/', '-_'), '='); // base64-URL-encoding
+    }
+
+    protected function getCodeChallengeMethod()
```
driesvints, line 481: We'll need DocBlocks for all these new methods.

src/Two/AbstractProvider.php
```diff
@@ -434,6 +456,32 @@ protected function getState()
...
+    protected function getCodeVerifier()
+    {
+        return Str::random(96);
+    }
+
+    protected function getCodeChallenge() 
+    {
+        $verifier = $this->request->session()->pull('code_verifier');
+        $hashed = hash('sha256', $verifier, true);
+        return rtrim(strtr(base64_encode($hashed), '+/', '-_'), '='); // base64-URL-encoding
```
driesvints, line 478:
```suggestion

        return rtrim(strtr(base64_encode($hashed), '+/', '-_'), '=');
```

tests/Fixtures/OAuthTwoWithPKCETestProviderStub.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Laravel\Socialite\Tests\Fixtures;
+
+class OAuthTwoWithPKCETestProviderStub extends OAuthTwoTestProviderStub 
+{
+	protected $usesPKCE = true;
```
driesvints, line 7:
```suggestion
    protected $usesPKCE = true;
```

tests/OAuthTwoTest.php
```diff
@@ -24,17 +25,67 @@ protected function tearDown(): void
...
 
-    public function testRedirectGeneratesTheProperIlluminateRedirectResponse()
+    public function testRedirectGeneratesTheProperIlluminateRedirectResponseWithoutPKCE()
     {
         $request = Request::create('foo');
         $request->setLaravelSession($session = m::mock(Session::class));
-        $session->shouldReceive('put')->once();
+
+        $state = null;
+        $closure = function($name, $stateInput) use(&$state) {
```
driesvints, line 34:
```suggestion
        $closure = function($name, $stateInput) use (&$state) {
```

tests/OAuthTwoTest.php
```diff
@@ -24,17 +25,67 @@ protected function tearDown(): void
...
+
+    private static $codeVerifier = null;
+
+    public function testRedirectGeneratesTheProperIlluminateRedirectResponseWithPKCE()
+    {
+        $request = Request::create('foo');
+        $request->setLaravelSession($session = m::mock(Session::class));
+
+        $state = null;
+        $closure = function($name, $value) use(&$state, &$codeVerifier) {
```
driesvints, line 60:
```suggestion
        $closure = function($name, $value) use (&$state, &$codeVerifier) {
```

tests/OAuthTwoTest.php
```diff
@@ -24,17 +25,67 @@ protected function tearDown(): void
...
-        $session->shouldReceive('put')->once();
+
+        $state = null;
+        $closure = function($name, $stateInput) use(&$state) {
+            if($name == 'state') {
+                $state = $stateInput;
+                return true;
+            } else {
+                return false;
+            }
```
driesvints, line 40:
```suggestion
            if ($name === 'state') {
                $state = $stateInput;

                return true;
            }

            return false;
```

tests/OAuthTwoTest.php
```diff
@@ -24,17 +25,67 @@ protected function tearDown(): void
...
+        $closure = function($name, $value) use(&$state, &$codeVerifier) {
+            if($name == 'state') {
+                $state = $value;
+                return true;
+            } elseif($name == 'code_verifier') {
+                self::$codeVerifier = $value;
+                return true;
+            } else {
+                return false;
+            }
```
driesvints, line 69:
```suggestion
            if ($name === 'state') {
                $state = $value;

                return true;
            } elseif ($name === 'code_verifier') {
                self::$codeVerifier = $value;

                return true;
            }

            return false;
```

tests/OAuthTwoTest.php
```diff
@@ -24,17 +25,67 @@ protected function tearDown(): void
...
+        };
+
+        $codeVerifierClosure = function($name) {
+            if($name == 'code_verifier') {
```
driesvints, line 73:
```suggestion
            if ($name === 'code_verifier') {
```

driesvints (changes requested):

tests/OAuthTwoTest.php
driesvints: You should be able to move this into the test method itself like this:

```php
public function testRedirectGeneratesTheProperIlluminateRedirectResponseWithPKCE()
{
    static $codeVerifier = null;
```

aaronpk: Strangely that isn't working. The first closure sets that value, but then it appears to be null by the time the second closure goes to read it.

src/Two/AbstractProvider.php
```diff
@@ -284,13 +300,19 @@ public function getAccessTokenResponse($code)
...
+        $fields = [
             'grant_type' => 'authorization_code',
             'client_id' => $this->clientId,
             'client_secret' => $this->clientSecret,
             'code' => $code,
             'redirect_uri' => $this->redirectUrl,
         ];
+
+        if ($this->usesPKCE()) {
+            $field['code_verifier'] = $this->request->session()->pull('code_verifier');
```
wietsewarendorff, line 312:
```suggestion
            $fields['code_verifier'] = $this->request->session()->pull('code_verifier');
```

src/Two/AbstractProvider.php
```diff
@@ -434,6 +456,49 @@ protected function getState()
...
+    }
+
+    /**
+     * Generates the PKCE code challenge based on the PKCE code verifier in the session.
+     *
+     * @return string
+     */
+    protected function getCodeChallenge()
+    {
+        $verifier = $this->request->session()->pull('code_verifier');
```
wietsewarendorff, line 486: If you pull the value from the session here, the `code_verifier` is no longer available for validation (on line 312)

driesvints:
Thanks @aaronpk. A PR to the docs is appreciated as well 🙂

ijansch:
@taylorotwell please check the above comments, the merged version contains a couple of typo's as per @wietsewarendorff's comment.

driesvints:
@aaronpk could you maybe send in a follow up pr for those typos? Thanks for mentioning that @ijansch

src/Two/AbstractProvider.php
aaronpk: Thanks, can't believe I missed that

src/Two/AbstractProvider.php
aaronpk: Thanks, I changed it to a `get`

aaronpk:
Here's the fix: #520

driesvints:
Follow up for stateless apps here: https://github.com/laravel/socialite/pull/523

Issue #522, opened 2021-02-18
v5.2.0 generate fatal error when used in REST endpoints


aprokopenko:
- Socialite Version: 5.2.0
- Laravel Version: 8.28.1
- PHP Version: 7.4.12
- Database Driver & Version: mysql driver, MySQL 8 

### Description:
We have a stateless application, which provides the only API enpoints. This means we don't have any sessions inside.
After update to 5.2.0 and using Google auth we now get the fatal error on the endpoint, which returns auth URLs for social networks.

Laravel fatal response:
![image](https://user-images.githubusercontent.com/1842666/108336735-088b7380-71dd-11eb-9f40-e34ac9e30b8c.png)

Sentry debug backtrace:
![image](https://user-images.githubusercontent.com/1842666/108336971-48525b00-71dd-11eb-97c5-3dc97bfad111.png)

### Steps To Reproduce:
- Create REST application, without session
- Try to get Google redirect URL

driesvints:
This package is only meant for web based sessions and not API authentication.

aprokopenko:
Previous version 5.1.3 worked great.

aprokopenko:
I think the correct behaviour will be to add config for enabling PCKE if it is supported by any provider.

In this case anyone can use this package

driesvints:
@aprokopenko PCKE support was added in the latest release: https://github.com/laravel/socialite/pull/518

driesvints:
I now see more clearly what you mean. @aaronpk do you know why your PR has broken things for use cases like @aprokopenko's one?

aprokopenko:
Yes, and this new addon for PCKE generates a fatal error because you use a session to store code_verifier and code_challenge. In the API you don't need this, because codes client secret/key are in safe (you can't get codes like in mobile APP or SPA). So PCKE is not needed in this case.

So more friendly is to have config for using PCKE, it can be ON by default, but should be the ability to turn it off.

I can't even override anything, because you create providers as `new $provider` and I can't bind my own class and set pcke property to false to make it works again.

So, unfortunately, I need to downgrade right now and stop updating this package to the latest versions.

driesvints:
I agree, it should have been opt-in. Not on by default. That should have been done in a new major version.

aprokopenko:
Cool, thanks.

driesvints:
[...]

PR #523, opened 2021-02-18
[5.x] Make PKCE opt-in

driesvints:
This makes https://github.com/laravel/socialite/pull/518 opt-in since it seems to be breaking apps using the Google provider in API's.

Fixes https://github.com/laravel/socialite/issues/522

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -467,6 +467,18 @@ abstract class AbstractProvider implements ProviderContract
     }
 
     /**
+     * Enables PKCE for the provider.
+     *
+     * @return $this
+     */
+    protected function enablePKCE()
+    {
+        $this->usesPKCE = true;
+
+        return $this;
+    }
+
+    /**
      * Generates a random string of the right length for the PKCE code verifier.
      *
      * @return string
--- a/src/Two/GoogleProvider.php
+++ b/src/Two/GoogleProvider.php
@@ -25,13 +25,6 @@ class GoogleProvider extends AbstractProvider implements ProviderInterface
     ];
 
     /**
-     * Indicates if PKCE should be used.
-     *
-     * @var bool
-     */
-    protected $usesPKCE = true;
-
-    /**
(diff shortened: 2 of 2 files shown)


aaronpk:
See my comments in #518 before merging this. Stateless apps can't use the OAuth authorization flow at all it requires maintaining application state in order to check the OAuth `state` parameter as well as the new PKCE parameters.

driesvints:
Converting to draft until we've gotten to the bottom of this.

driesvints:
@aaronpk I'm going to go ahead with this PR by disabling PKCE by default and making it opt-in since this seems to break apps using the Google Provider in an API: https://github.com/laravel/socialite/issues/524 https://github.com/laravel/socialite/issues/522

Until we've fully figured out what's causing this we should keep it opt-in.

aaronpk:
Okay I've reviewed everything again, and I'm guessing that people who are using this from a stateless API are just leaving out the OAuth `state` check entirely, since the `hasInvalidState` checks for that `stateless` property before checking the OAuth state value. Omitting the `state` check opens up CSRF attacks on the client, which was the reason for including the `state` parameter in OAuth in the first place. That said, PKCE prevents CSRF attacks in a cleaner way, by having the server enforce it. (it also prevents some other attacks as well).

The best way to resolve this in a future update would be to provide two ways for people to add PKCE when using it from an API without a session, either:
* make the `codeVerifier` that the library generates available to the calling code so that it can provide it to the `getAccessTokenResponse` method
* set the `codeVerifier` value when they make the method call to get the authURL, and also provide the `codeVerifier` value to the `getAccessTokenResponse` method

driesvints:
@aaronpk think it's best that you just attempt a PR with what you want to do so we can look at actual code.

binotaliu:
`enablePKCE()` method is declared as `protected`. How do I enable it without extending my own provider?

Lukerayn3r:
@aaronpk Did you ever get chance to work on a PR for using PKCE on a stateless API?

PR #537, opened 2021-03-23
Make getAuthUrl public

### LINKED ISSUE #536: Allow returning auth url for stateless SPA
I'm trying to handle social logins from a SPA frontend.

My plan is fetching auth url from frontend so redirect and callback can be handled from there.

However, `getAuthUlrl` is marked as protected. 

Is there another way of doing this?
(I don't want to redirect to backend URL so users get redirected again to google)

```
    public function redirect($provider)
    {
        return Socialite::driver($provider)->stateless()->getAuthUrl();
    }

```

eduardoarandah:
When using a SPA, you might want to redirect from your frontend.

With this change you can now get the URL for redirection like this:

```
return Socialite::driver($provider)
    ->stateless()
    ->getAuthUrl(null);

```

Fixes #536

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -127,7 +127,7 @@ abstract class AbstractProvider implements ProviderContract
      * @param  string  $state
      * @return string
      */
-    abstract protected function getAuthUrl($state);
+    abstract public function getAuthUrl($state);
 
     /**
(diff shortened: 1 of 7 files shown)


driesvints:
This seems to break the test suite and we can't change the method visibility on a stable release. Please try master with passing tests.

PR #538, opened 2021-03-24
[6.x] Make getAuthUrl public

eduardoarandah:
When using a SPA, you might want to redirect from your frontend.

With this change you can now get the URL for redirection like this:

```
return Socialite::driver($provider)
    ->stateless()
    ->getAuthUrl(null);

```

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -127,7 +127,7 @@ abstract class AbstractProvider implements ProviderContract
      * @param  string  $state
      * @return string
      */
-    abstract protected function getAuthUrl($state);
+    abstract public function getAuthUrl($state);
 
     /**
      * Get the token URL for the provider.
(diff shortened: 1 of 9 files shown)


eduardoarandah:
@driesvints Please tell me if this test with `->stateless()` makes any difference or should I remove it

https://github.com/laravel/socialite/pull/538/commits/92b5506d06a8f899622644020218361e11c0332b#diff-281fc2076589734986e27d83aadcb4f46089dd5f4a8bc0c978532171426b6a23R190

driesvints:
@eduardoarandah it's fine. Although I'm still in doubt on wether this is needed at all.

eduardoarandah:
@driesvints I guess it might not be needed for sites with Inertia, given that routing is made in laravel. 

This is the problem I want to solve:

I'm thinking about SPA sites with frontend routing. 

You have a button 
[ login with google ]

The SPA needs to know the URL to send the user on click. 

For that, it can call the backend (which holds the keys in its .env) in a route `api/oauth/login-url/{provider}`

WIth current implementation I guess it's possible to do this in backend, to tell frontend what's the social URL. 
But it's not elegant, why generate a response?

```
$response = Socialite::driver($provider)->stateless()->redirect();
return $response->getTargetUrl();
```

Is this attempted solution alright? what do others do with their SPA apps?

PR #544, opened 2021-06-09
Adding userFromToken to Provider Interface

OsamaMohammed:
The Provider interface should have `userFromToken` function in it to make it visible for code snippets and fix the annoying IDE error warning 'function not found'.


driesvints:
We can't add methods to contracts on stable releases. Please send to master.

PR #545, opened 2021-06-10
Adding userFromToken to Provider Interface

OsamaMohammed:
The Provider interface should have `userFromToken` function in it to make it visible for code snippets and fix the annoying IDE error warning 'function not found'.


driesvints:
This isn't necessary for Socialite's provider contract. This is a method for a concrete implementation. Some drivers might not need this method to function. You need to type-hint the abstract provider class or the concrete driver provider class.

PR #548, opened 2021-08-04
Handle 'scope' for Twitter Oauth1

Diegslapasteque:
Hey,
I recently [made a pull request](https://github.com/thephpleague/oauth1-client/pull/138) in the [thephpleague/oauth1-client](https://github.com/thephpleague/oauth1-client) repository to handle the _scope_ for Twitter Oauth1 (with the parameter _x_auth_access_type_).
I use socialite in my app and I need to be able to set this scope manually, so I added a new method in the TwitterProvider.php to handle it.

### DIFF
--- a/src/One/TwitterProvider.php
+++ b/src/One/TwitterProvider.php
@@ -32,4 +32,15 @@ class TwitterProvider extends AbstractProvider
             'avatar_original' => str_replace('_normal', '', $user->imageUrl),
         ]);
     }
+
+    /**
+     * Overrides the access level an application requests to a users account.
+     *
+     * @param string $scope
+     * @return void
+     */
+    public function scope(string $scope)
+    {
+        $this->server->setApplicationScope($scope);
+    }
 }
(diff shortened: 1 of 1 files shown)


ZekLouis:
Same issue here, this fix would be really useful ! Thanks @Diegslapasteque

taylorotwell:
Can you provide an example of how it would be used?

Diegslapasteque:
### I will take the use case for our app as an example
For all our users, we just need to get a Twitter oauth token with the default _read_ access. We prefer not to use the _write_ access by default to ask the minimum permissions to avoid scaring our users on the Twitter oauth page (permission to publish tweets, etc.).

So in our config/services.php, we have this :
```
'twitter' => [
    'client_id' => env('TWITTER_CLIENT_ID'),
    'client_secret' => env('TWITTER_CLIENT_SECRET'),
    'redirect' => env('TWITTER_REDIRECT'),
    'scope' => 'read'
]
```

But we have some features in our app that require a _write_ access from the Twitter API. When a user try to access one of them, he is going to be redirected to the Twitter oauth again, this time with a _write_ access.
In this case, the method `setApplicationScope()` that I added will be useful to change the access level.
```
/** @var TwitterProvider $driver */
$driver = Socialite::driver('twitter');
$driver->setApplicationScope('write');
$driver->redirect();
```

Is it okay for you?

PR #554, opened 2021-10-04
[5.x] Don't use strlen to check $state

hrs-o:
If $state is null, PHP8.1 will throw a warning with strlen.
`strlen(): Passing null to parameter #1 ($string) of type string is deprecated`.

https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg


driesvints:
Just doing the below is probably easier:

```php
return ! (strlen($state ?? '') > 0 && $this->request->input('state') === $state);
```

PR #565, opened 2021-11-28
Fixed DocBlock return type

shehi:
Need this DocBlock fix so that IDEs didn't complain when we type-hint our extensions and implementations.

### DIFF
--- a/src/Contracts/Provider.php
+++ b/src/Contracts/Provider.php
@@ -7,7 +7,7 @@ interface Provider
     /**
      * Redirect the user to the authentication page for the provider.
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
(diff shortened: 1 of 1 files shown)


src/Contracts/Provider.php
```diff
@@ -7,7 +7,7 @@ interface Provider
     /**
      * Redirect the user to the authentication page for the provider.
      *
-     * @return \Illuminate\Http\RedirectResponse
+     * @return \Symfony\Component\HttpFoundation\RedirectResponse|\Illuminate\Http\RedirectResponse
```
shehi, line 10: @taylorotwell , not to complain, but `\Illuminate\Http\RedirectResponse` already extends `\Symfony\Component\HttpFoundation\RedirectResponse`. Why put both of them here?! :)

Issue #568, opened 2021-12-03
Set the scopes on the Socialite\User instance


vanyamil:
Hello,

I am currently working on mixing custom Google API calls alongside the existing Socialite implementation. As part of requesting additional scopes (using Google's [incremental authorization](https://developers.google.com/identity/protocols/oauth2/web-server#incrementalAuth)), I would like the possibility of seeing all scopes given by the current access token. These are returned in the ->getAccessTokenResponse() method under the `scope` method.

Since this is part of the guidelines on the access token response ([OAuth2 simplified](https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/)), I think it would be beneficial to add this to the whole AbstractProvider class. I'm not good at PRs so some ideas here:

In Two/AbstractProvider.php@user, before the return, add 
```php
if(Arr::exists($response, 'scope'))
    $this->user->setScopes(explode($this->scopeSeparator, Arr::get($response, 'scope')))
```
In Two/User.php, add a new method and field:

```php
    /**
     * The user's granted scopes.
     *
     * @var array
     */
    public $scopes;

    /**
     * Set the scopes that the access token has permission for.
     *
     * @param  array $scopes
     * @return $this
     */
    public function setScopes($scopes)
    {
        $this->scopes = $scopes;

        return $this;
    }
```

The only missing part of the puzzle is how to track which scopes should be set if the scopes match what was requested and no scopes field is placed. It looks like we'd have to insert the requested scopes into the session and pull() them like the state? But maybe I'm not seeing the correct approach here

Thanks!

vanyamil:
If we do store the scopes into session, then we won't even need the `Arr::exists` check - just set the default value on the `Arr::get` to be the `$this->request->session()->pull('scopes')`

driesvints:
Heya, I'm sorry but we currently do not have plans to work on something like this ourselves. You can always attempt a PR or look around for someone willing to co-create one. Thanks

PR #572, opened 2022-01-18
add/set approvedScopes in User

lsnider:
It's possible for a user to grant a subset of the requested scopes.  Google has been pushing this recently...see the attached screenshot of their latest Oauth page.  Their latest default is to have no scopes checked making it quite easy for a user to select a subset of the requested scopes.

Part of the Oauth flow suggests the providers should return the approved scopes as part of the `code -> access_token` exchange if the user auth'ed a subset of the requested scopes. In my testing of Google, M365, Slack, and Zoom, the `scope` field is always returned as part of the `code -> access_token` exchange.  

There was a [recent issue ](https://github.com/laravel/socialite/issues/568) asking for this feature.

As pointed out in the link within [issue#568](https://github.com/laravel/socialite/issues/568), we should expect `code` -> `access_token`, `refresh_token (optional)`, `expires_at`, `scope (optional)`
 
There was a [recent PR ](https://github.com/laravel/socialite/pull/152/files) which brought the `User` class up from just `access_token` to also include `refresh_token`, and `expires_at`.

This PR also brings in `scope` to the `User` object.  That would complete the `User` object for storing everything the `code -> access_token` exchange returns. 

Would you like to merge this up stream?  If so, I can do whatever I can to make this acceptable up steam.  I tested this with the providers we are using.  I can also put in a PR for the SocialiteProviders manager.  In SocialiteProviders, it's much like this PR but we set the scopes [here](https://github.com/SocialiteProviders/Manager/blob/master/src/OAuth2/AbstractProvider.php#L54) and add a `parseApprovedScopes` function like the others in that file.

Tested with native Oauth Provider:
- Google Personal
- Google Workspace/Gsuite

Tested with SocialiteProviders (needs a simple PR):
- M365
- Slack (we have a fork of the Slack Provider)

Google Oauth Screen

If it's helpful, here are a few documents showing the `scope` field gets returned from a auth code to access token exchange

- [GitHub](https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#2-users-are-redirected-back-to-your-site-by-github)
- [Google](https://developers.google.com/identity/protocols/oauth2/web-server#exchange-authorization-code)
- [M365](https://docs.microsoft.com/en-us/graph/auth-v2-user#response)
- [Slack](https://api.slack.com/legacy/oauth#authenticating-users-with-oauth__the-oauth-flow__step-3---exchanging-a-verification-code-for-an-access-token)
- [Zoom](https://marketplace.zoom.us/docs/guides/auth/oauth#step-2-request-access-token)

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -245,7 +245,7 @@ abstract class AbstractProvider implements ProviderContract
         return $this->user->setToken($token)
                     ->setRefreshToken(Arr::get($response, 'refresh_token'))
                     ->setExpiresIn(Arr::get($response, 'expires_in'))
-                    ->setApprovedScopes(Arr::get($response, 'scope'));
+                    ->setApprovedScopes(explode(' ', Arr::get($response, 'scope', '')));
     }
 
     /**
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -245,7 +245,7 @@ abstract class AbstractProvider implements ProviderContract
         return $this->user->setToken($token)
                     ->setRefreshToken(Arr::get($response, 'refresh_token'))
                     ->setExpiresIn(Arr::get($response, 'expires_in'))
-                    ->setApprovedScopes(explode(' ', Arr::get($response, 'scope', '')));
+                    ->setApprovedScopes(explode($this->scopeSeparator, Arr::get($response, 'scope', '')));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't seem to have permission to edit this PR?

lsnider:
@taylorotwell Were you just having trouble marking it as a draft?  I don't see any permissions controls on my side.

lsnider:
If this PR gets merged, here is the corresponding [PR for SocialliteProviders](https://github.com/SocialiteProviders/Manager/pull/206)

Issue #570, opened 2022-01-12
Twitter OAuth 2.0


NewBuidler:
Hi,

It seems like Twitter is using the OAuth 2.0 now. 
Is there any plan to use it in Socialite, too? 
Thank you.

driesvints:
Yeah it seems we can look into this. I can look into this eventually but of course we'd accept PR's for anyone wanting to help out here.

https://twittercommunity.com/t/announcing-oauth-2-0-general-availability/163555/2

PR #574, opened 2022-01-23
[5.x] Add Twitter OAuth2 provider

dehbka:
Unfortunately, at the moment the endpoint for receiving email (https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials) has not been added to the twitter v2 api.

https://developer.twitter.com/en/docs/twitter-api/migrate/twitter-api-endpoint-map

for: https://github.com/laravel/socialite/issues/570

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -13,6 +13,7 @@ use Laravel\Socialite\Two\GithubProvider;
 use Laravel\Socialite\Two\GitlabProvider;
 use Laravel\Socialite\Two\GoogleProvider;
 use Laravel\Socialite\Two\LinkedInProvider;
+use Laravel\Socialite\Two\TwitterProvider as TwitterOAuth2Provider;
 use League\OAuth1\Client\Server\Twitter as TwitterServer;
 
 class SocialiteManager extends Manager implements Contracts\Factory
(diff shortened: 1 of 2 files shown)


driesvints (changes requested):

taylorotwell:
Converting to draft while Dries' requested changes are made.

taylorotwell:
@dehbka so, how would you actually use this? `Socialite::driver('twitter-oauth-2')`?

dehbka:
yeap

> @dehbka so, how would you actually use this? `Socialite::driver('twitter-oauth-2')`?

michaeldyrynda:
@dehbka did you run into any issues with this?

Seems as though if you have a confidential client (web app or automated bot), you _also_ need to send a basic auth header with the client ID and secret as username and password. 

Not doing so results in authorisation failures when attempting to get a oauth token/fetching the authenticated user. 

It does seem as though you can send the authorisation head for either public or confidential clients, so it may be sufficient to override the `getAccessTokenResponse` method on the `TwitterProvider`  

Happy to send a PR for this, I spent longer than I care to admit trying to figure out why it wasn’t working 🥲

ahmedsayedabdelsalam:
I know it's not related to laravel it might be permission or something, but i have little question if anyone can help
I was using 'twitter' driver for oauth1 and switched to 'twitter-oauth-2' but the user email now returns 'null'.

PR #581, opened 2022-04-12
Add `refreshToken` method for OAuth 2.0 providers

viktorruskai:
## Description
This PR contains a method to refresh token received from social networks  via OAuth2. This method will be helpful for projects that are storing access tokens (also refresh tokens) and they are using them on behalf of the users to receive some data, metrics etc.

Sometimes these kinds of tasks are done by cron in the night and without acknowledging users. With this feature programmers can implement the case when a users' access tokens are going to expire and they don't want to reauthorize users.

## Usage example
Users' access tokens would be stored in `user_tokens` table (model `UserToken`). The table consists of `id, user_id, network, access_token, refresh_token, expires_at, updated_at, created_at` columns.

**app/Commands/Console/RefreshTokenCommand.php**
```php
class RefreshFacebookTokens extends Command
{

    protected $signature = 'token:refresh';
    protected $description = 'Refresh tokens for all supported social networks.';

    /**
     * Execute the console command.
     */
    public function handle(): int
    {
        $userTokens = UserTokens::where('expires_at', '<=', Carbon::now())->get();

        foreach ($userTokens as $userToken) {
            $token = Socialite::driver($userToken->social_network)->refreshToken($userToken->refresh_token ?? $userToken->access_token); // (*)

            if ($token['access_token']) {
                $userToken->access_token = $token['access_token'];
            }

            $userToken->refresh_token = $token['refresh_token'] ?? null;
            $userToken->expires_at = Carbon::now()->addSeconds($token['expires_in']);
            $userToken->save();
        }

        return 0;
    }
}
```

_(*) Facebook does not return `refresh_token`. Facebook uses `access_token` in order to receive Long-Lived token._

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -103,7 +103,7 @@ abstract class AbstractProvider implements ProviderContract
     protected $user;
 
     /**
-     * Grant type for refresh token
+     * Grant type for refresh token.
      *
      * @var string
      */
@@ -269,7 +269,7 @@ abstract class AbstractProvider implements ProviderContract
     }
 
     /**
-     * Refresh User's token
+     * Refresh User's token.
      *
      * @param  string  $token
      * @param  bool  $useScopes
@@ -356,7 +356,7 @@ abstract class AbstractProvider implements ProviderContract
             $this->refreshTokenGrantType => $token,
         ];
 
-        if ($useScopes && !empty($this->getScopes())) {
+        if ($useScopes && ! empty($this->getScopes())) {
             $fields['scope'] = $this->formatScopes($this->getScopes(), $this->scopeSeparator);
         }
 
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -103,6 +103,13 @@ abstract class AbstractProvider implements ProviderContract
     protected $user;
 
     /**
+     * Grant type for refresh token
+     *
+     * @var string
+     */
+    protected $refreshTokenGrantType = 'refresh_token';
+
+    /**
      * Create a new provider instance.
      *
      * @param  \Illuminate\Http\Request  $request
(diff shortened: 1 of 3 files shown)


src/Two/AbstractProvider.php
```diff
@@ -261,6 +268,23 @@ public function userFromToken($token)
...
+     * Refresh User's token.
+     *
+     * @param  string  $token
+     * @param  bool  $useScopes
+     * @return array
+     */
+    public function refreshToken($token, $useScopes = false)
+    {
+        $response = $this->getHttpClient()->post($this->getTokenUrl(), [
+            'headers' => ['Content-type' => 'application/x-www-form-urlencoded'],
```
driesvints, line 281: Wondering if this is also needed for all other OAuth 2.0 drivers. Did you only test Facebook? Can you check into a few more drivers?

viktorruskai: Yes, I tested it for Facebook, Google, LinkedIn (LinkedIn App must be partner with LinkedIn), Gitlab and Github. It worked correctly.

taylorotwell:
Generally would expect the token to be "refreshed" and updated in your database each time a user logs into your application? Regardless, a bit concerned about the scope situation here. What if they are providing scopes at runtime for the provider? This would just use the default scopes that Socialite specifies on the provider.

viktorruskai:
@taylorotwell my answers for your questions:

> Generally would expect the token to be "refreshed" and updated in your database each time a user logs into your application?

- Yes, you are right, but what about applications that need access tokens to be ready for cron (users are not logged in, but their access tokens / refresh tokens are stored in database) -> Cron job needs valid access token for social media requests.
- For example, Facebook access token is valid for 2 hours, after this period this action would be invalid.

> Regardless, a bit concerned about the scope situation here. What if they are providing scopes at runtime for the provider? This would just use the default scopes that Socialite specifies on the provider.

- When refreshing token, scopes are [optional](https://datatracker.ietf.org/doc/html/rfc6749#section-6), that's why `refreshToken` method has second parameter `$useScopes` set to `false` (by default), so the scopes are not in the request and the authorization server should not take care of scopes (issued).

- In case when applications would need to use specific scopes, which were issued during the authorization process (those scopes must be the same in the refresh process):
```php
Socialite::driver('github')->scopes(['read:user', 'read:email', ...])->refreshToken('...', true);
```

- In scenario when scopes were changed by provider the authorization server should return error _(or issue new access token)_ during refresh process.

PR #588, opened 2022-05-06
Update FacebookProvider.php

volkankaban:


driesvints:
There's no explanation here and tests are failing.

Issue #595, opened 2022-07-17
`twitter-oauth-2` config key issue


plabbett:
- Socialite Version: 5.5.2
- Laravel Version: 9.20.0
- PHP Version: 8.1.8
- Database Driver & Version:  Eloquent/mysql via sail

### Description:

I'm not sure if this is a documentation issue or a behavior issue. The documentation for [Laravel Socialite](https://github.com/laravel/docs/blob/9.x/socialite.md) explains:

> These credentials should be placed in your application's `config/services.php` configuration file, and should use the key `facebook`, `twitter` (OAuth 1.0), `twitter-oauth-2` (OAuth 2.0), `linkedin`, `google`, `github`, `gitlab`, or `bitbucket`, depending on the providers your application requires:

However, `twitter-oauth-2` does not seem to work unless you rename the services key from `twitter-oauth-2` to `twitter` (or put the `client_id` and `client_secret` values into existing `twitter` key.) 

I have tried `php artisan config:clear` and `sail artisan config:clear` and the only way I could it get things going with OAuth2 was to use the `twitter` key in `config/services.php` for the `twitter-oauth-2` integration. 

Here is the Flare Share: https://flareapp.io/share/LPdozlQm

### Steps To Reproduce:

1. `composer create-project --prefer-dist laravel/laravel ./folder`
2. `cd folder`
3. `composer require laravel/socialite`
4. Configure `config/services.php` with a key for `twitter-oauth-2` and the required `client_id`, `client_secret`, `redirect` keys.
5. Try to login using `twitter-oauth-2`

I'm wondering if [adjusting this would be a fix](https://github.com/laravel/socialite/blob/89f50bc8a279ca5acedbd7bb494e690bb22bc8b8/src/SocialiteManager.php#L141) or if this should go back to the Laravel docs?

```php
 $config = $this->config->get('services.twitter') ?? $this->config->get('services.twitter-oauth-2');
```

driesvints:
Actually using the `twitter` key will just work fine but I can understand that the docs are a bit confusing. I've sent in a PR with your suggestion. Thanks!

PR #596, opened 2022-07-18
Add Twitter OAuth 2 fallback call

driesvints:
This allows the Twitter OAuth 2 driver to fallback on the `twitter-oauth-2` config key as explained in the docs.

See https://github.com/laravel/socialite/issues/595


tehwave:
@dependabot squash and merge

PR #606, opened 2022-09-27
Added types. Fixed consistency. Updated composer and GitHub actions

SuperDJ:
- Added types where possible
- Fixed test name consistency
- Updated composer package versions and GitHub test actions accordingly

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,27 +13,14 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [7.2, 7.3, 7.4, 8.0, 8.1]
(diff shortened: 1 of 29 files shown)


driesvints:
Thanks but sorry, this is way too much for us to review. We don't use most of the things that got changed here.

SuperDJ:
@driesvints would you possibly accept it if there are multiple smaller PRs? 
Could you clarify what you mean with the following?
>We don't use most of the things that got changed here.

driesvints:
Right now we don't have plans on doing any of these changes, sorry.

SuperDJ:
@driesvints is this still not usable for Laravel 10?

driesvints:
No, we have no plans to add types to our codebases. Only on stubs.

PR #607, opened 2022-10-28
Test on PHP 8.2

driesvints:
I've had to implement an attribute map on the abstract user entity to make sure dynamic property handling still worked as before.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,6 +2,9 @@ name: tests
 
 on:
   push:
+    branches:
+      - master
+      - '*.x'
   pull_request:
   schedule:
     - cron: '0 0 * * *'
@@ -13,7 +16,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [7.2, 7.3, 7.4, 8.0, 8.1]
+        php: [7.2, 7.3, 7.4, 8.0, 8.1, 8.2]
         laravel: [^6.0, ^7.0, ^8.0, ^9.0]
         exclude:
           - php: 7.2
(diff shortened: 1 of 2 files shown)


sebastiaanluca:
Sorry I'm always the bringer of bad news 😄

Coming from v5.5.5, caught a bug in our implementation of Socialite where `token_secret` would now be `null` if you get it via the `object_get()` helper:
```
-  'token_secret' => 'abc',
+  'token_secret' => NULL,
```

My findings:

```php
// abc
$socialiteUser->tokenSecret,

// abc
data_get($socialiteUser, 'tokenSecret'),

// null
object_get($socialiteUser, 'tokenSecret'),

// ZOt24fxKTmkxYRBnUOgui8BYdg6TzsUn7deYgByZMJmAqMN9qTswRRiSVeCk
$socialiteUser->token,

// ZOt24fxKTmkxYRBnUOgui8BYdg6TzsUn7deYgByZMJmAqMN9qTswRRiSVeCk
data_get($socialiteUser, 'token'),

// ZOt24fxKTmkxYRBnUOgui8BYdg6TzsUn7deYgByZMJmAqMN9qTswRRiSVeCk
object_get($socialiteUser, 'token'),
```

The weird thing here is it only fails for `tokenSecret` 🤨  Both are only defined in `\Laravel\Socialite\One\User` (not the parent `AbstractUser`) and are set in the same way. Have no idea at the moment what the difference is and why one returns `null`.

driesvints:
@sebastiaanluca heya, yes I suppose that's correct. Sorry you got bit here (again). But seems you're good and we don't need to rollback?

sebastiaanluca:
@driesvints I think you're good. The OAuth1 user has the `tokenSecret` property, so it should be accessible via `object_get()`. It's a real edge case 😄

PR #614, opened 2023-01-04
[5.x] Fix User return types

antoinelame:
The PR fixes the return types of a few `Laravel\Socialite\Contracts\User`'s methods. It will sove false positives when using static analyzers.


driesvints:
Fyi, `?string` is a PHP type thing, not a valid DocBlock type. Needs to be `string|null` instead. I adjusted this.

PR #617, opened 2023-01-12
Fix PKCE causing session reach with stateless

z-siddy:
Code improvement for PKCE usage while using `stateless` method in `TwitterProvider`. Currently the `stateless` method does not prevent the default enabled property `$usesPKCE` in `TwitterProvider.php` and checking if the driver uses state first resolves unnecessary PKCE usage (which also uses sessions).

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -164,12 +164,12 @@ abstract class AbstractProvider implements ProviderContract
 
         if ($this->usesState()) {
             $this->request->session()->put('state', $state = $this->getState());
+            
+            if ($this->usesPKCE()) {
+                $this->request->session()->put('code_verifier', $this->getCodeVerifier());
+            }
         }
-
-        if ($this->usesPKCE()) {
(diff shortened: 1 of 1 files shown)


driesvints:
I thought this was placed outside the state check for a reason but cannot remember why. @aaronpk do you maybe know?

z-siddy:
> I thought this was placed outside the state check for a reason but cannot remember why. @aaronpk do you maybe know?

Well, I looked up Twitter's docs after unsuccessful login attempts and came accross this:
https://developer.twitter.com/en/docs/authentication/oauth-2-0/authorization-code

It seems that they require PKCE with OAuth 2.0, so I guess this PR will not solve the issue.

z-siddy:
Not package related, but it seems that if you consume Laravel's API routes and apply `\Illuminate\Session\Middleware\StartSession::class` middleware, you can consume the session with stateless method.

taylorotwell:
Closing this as it sounds like it's not needed or doesn't fix the issue you're after? Correct me if I'm wrong.

PR #629, opened 2023-02-19
[5.x] Fix misspelling uppercase "GitHub"

datlechin:
Fix misspelling uppercase "GitHub"


driesvints:
This might be a breaking change for case sensitive filesystems.

PR #631, opened 2023-02-28
Add PHPUnit 10

SuperDJ:
Add PHPUnit 10


driesvints:
Heya, thank you for this but for now it's fine to just use the current PHPUnit versions.

Issue #634, opened 2023-04-14
New Google sign-in


mikelina:
On March 31, 2023 Google deprecated the JS Library used for sign-in (https://developers.google.com/identity/sign-in/web/sign-in?hl=en)
The new library (https://developers.google.com/identity/gsi/web/guides/overview?hl=en) returns a JWT token instead of an access token.
Will you update socialite in order to handle this new feature?

driesvints:
Heya, yeah I think we should migrate to the new way of doing things. This would have to be a PR sent in to master for the next major version.

The guide here can help: https://developers.google.com/identity/gsi/web/guides/migration?hl=en

We'd appreciate any help with this.

nunomaduro:
@mikelina Are you able to make a pull request fixing the issue you have described, along with corresponding tests?

CosminBd:
@nunomaduro I'm actively working on an implementation for this.

nunomaduro:
@CosminBd Thank you! Let's follow up on the discussion on the pull request itself! ✅

driesvints:
Looks like Taylor wants this to be a community driver, sorry.

mikield:
I do not think that this is an issue for [socialite](https://github.com/laravel/socialite). 
[socialite](https://github.com/laravel/socialite) is handling OAuth flow, but Google new auth system via Google Identify Service uses JWT auth, so there is no place to new google auth system for [socialite](https://github.com/laravel/socialite) package.

PR #636, opened 2023-05-13
[5.x] Add support for new google authentication library

CosminBd:
Created a new provider which uses the new Google Authentication Library (https://developers.google.com/identity/gsi/web/guides/overview?hl=en) 

This is related to #634.

To use the new library you would have to use `Socialite::driver('google-jwt')->user()`

the redirect method is deprecated, as google allows only for HTML and JS integrations, and now throws a message. 

For client side one could use:

 ```
<script src="https://accounts.google.com/gsi/client" async defer></script>

                    <div id="g_id_onload"
                         data-client_id="{{config('services.google.client_id')}}"
                         data-login_uri="{{config('services.google.redirect')}}"
                         data-auto_prompt="false">
                    </div>
                    <div class="g_id_signin"
                         data-type="standard"
                         data-size="large"
                         data-theme="outline"
                         data-text="sign_in_with"
                         data-shape="rectangular"
                         data-logo_alignment="left">
                    </div>
```

Basically, Google moved away from OAuth and now issue JWT's in order to authenticate the user.

My PR checks for the JWT, validates it and returns an User instance based on data provided.
I've also added a dependency for the openssl PHP extension, which is used to verify the signature of the JWT's with Google certs.

Both authentication methods are available via: 'google' and 'google-jwt' drivers.

I didn't add testing yet because I didn't know which package to use for mocking the JWT payloads.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
         "illuminate/contracts": "^6.0|^7.0|^8.0|^9.0|^10.0",
         "illuminate/http": "^6.0|^7.0|^8.0|^9.0|^10.0",
         "illuminate/support": "^6.0|^7.0|^8.0|^9.0|^10.0",
-        "league/oauth1-client": "^1.10.1"
+        "league/oauth1-client": "^1.10.1",
+        "ext-openssl": "*"
     },
     "require-dev": {
         "mockery/mockery": "^1.0",
--- a/src/Jwt/GoogleProvider.php
+++ b/src/Jwt/GoogleProvider.php
@@ -2,7 +2,6 @@
 
 namespace Laravel\Socialite\Jwt;
 
-use _PHPStan_a3459023a\Nette\Neon\Exception;
 use Illuminate\Support\Arr;
 use Illuminate\Http\Request;
 use Laravel\Socialite\Contracts\Provider;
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by nunomaduro** [comment]
@CosminBd Regarding your question on Twitter: By "Laravel coding standards", Dries only means that you should format the code exactly as typically is formatted on other places on the framework regarding PHP annotations, etc.  Here is an example:

```diff
class GoogleProvider implements Provider
{
+   /**
+    * Create a new google provider instance.
+    *
+    * @param  \Illuminate\Http\Request  $request
+    * @param  string  $clientId
+    */
    public function __construct(Request $request, $clientId)
    {
        $this->request = $request;
        $this->clientId = $clientId;
    }
```

Another example:

```diff
    /**
     * Verifies the signature of the JWT issued, via Google Certs
     *
     * @param  string  $jwt
+    * @return bool
-    * @return boolean
     */
-   protected function verifyJwtSignature($jwt){
+   protected function verifyJwtSignature($jwt)
+   {
-       $jwt_parts = explode('.', $jwt);
+       $jwtParts = explode('.', $jwt);
          // ...
```

nunomaduro:
@CosminBd Regarding testing, I would probably take a look about how currently things are being tested on other/existing providers, and try to follow the existing examples regarding testing.

taylorotwell:
Can be a community maintained driver.

codespearhead:
@CosminBd It seems that [SocialiteProviders/Providers](https://github.com/SocialiteProviders/Providers/issues/940#issuecomment-1858690586) is willing to accept it there.

Issue #638, opened 2023-05-29
Unable to use updated config object when using Laravel Octane


aprokopenko:
### Socialite Version

5.6.1

### Laravel Version

10.12.0

### PHP Version

8.2.6

### Database Driver & Version

MySQL 8.0.30 (official docker image)

### Description

When switching our project to Octane (v2) we found out, that Socialite package doesn't work.

We have RESTful API and provide social URLs with separate endpoint. Also we have dynamic redirect URLs with CSRF token inside to prevent some kind of attack. 

SocialiteManager::driver() method use config object like "services.{service}" to create drivers and 'redirect' key is mandatory. Before (with PHP-FPM) we just set `redirect` on each endpoint call with `config(['services.{service}.redirect' => $url])` and it works fine.
But after switching to Octane we get 500 error, that `redirect` key is underfined.

After deep debugging of this problem, we found out that Octane reset `SocialiteManager` on each request and assign new $container object to it. However, SocialiteManager is already booted and has `$this->config` property, which was loaded at the very beginning from previous requests and this object stays the same and contain outdated data.

So if I manipulate with config object of the Container, it doesn't affect the SocialiteManager `$this->config` and this leads to 500 error in our case.

I prepared a Pull Request to fix this. Would be great to release a hotfix with it.

### Steps To Reproduce

- Configure octane with socialite environment
- Do not configure driver in `services.php`
- Create some route to get oauth redirect url
- Set config() object for example for Google dynamically inside route (`config(['services.google' => $yourCredentials])`
- Try to get google driver with `Socialite::driver('google');`
- You will get fatal error that required keys are empty

(same scenario works fine with PHP-FPM)

aprokopenko:
Pull request: https://github.com/laravel/socialite/pull/639

nunomaduro:
Fixed: https://github.com/laravel/socialite/pull/639.

PR #639, opened 2023-05-29
Fix: Unable to use updated config object when using Laravel Octane

aprokopenko:
Fixing the issue described here: https://github.com/laravel/socialite/issues/638

Short summary:

Octane reset SocialiteManager on each request and assign new $container object to it. However, SocialiteManager is already booted and has $this->config property, which was loaded at the very beginning from previous requests and this object stays the same and contain outdated data.

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -222,6 +222,7 @@ class SocialiteManager extends Manager implements Contracts\Factory
     {
         $this->app = $container;
         $this->container = $container;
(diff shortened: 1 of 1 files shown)


nunomaduro (commented):
Lets wait for Taylor's opinion. However, in my opinion, the configuration of an Laravel project should never change during the request lifecycle.

aprokopenko:
Thanks for a quick merge. When could I expect for the next release of this package?

PR #645, opened 2023-07-12
Update Slack provider to use v2 API and allow Bot tokens

jbrooksuk:
This PR changes a few things with Slack:

1. Switches to the v2 API.
2. Adds a `asBotUser()` method that switches the type of scopes requested. Slack has two kinds of tokens; `xoxb` (bot) and `xoxp` (user). When using `asBotUser()` the `user()` method will return an empty user, but a bot token will still be available.

### DIFF
--- a/src/Two/SlackProvider.php
+++ b/src/Two/SlackProvider.php
@@ -22,6 +22,18 @@ class SlackProvider extends AbstractProvider
     protected $scopeKey = 'user_scope';
 
     /**
+     * Indicate that the requested token should be for a bot user.
+     *
+     * @return $this
+     */
+    public function asBotUser()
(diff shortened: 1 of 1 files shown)


driesvints (approved):

joedixon (approved):

ataldev:
We should be able to use bot user_scopes and scopes in the same redirect.
For example `user_scopes=identity.email&scopes=commands`
Currently, to achieve this, the asBotUser can not be utilized. And we should use something like:
```php
return Socialite::driver('slack')->scopes(['identity.email'])->with(['scopes' => 'commands'])->redirect()
```
Which is not ideal.

PR #648, opened 2023-07-28
[5.x] Add getToken() method to User contract

josepostiga:
This PR adds a new `getToken()` method to the abstract `User` class that returns the `$token` property value that both `\Laravel\Socialite\One\User` and `\Laravel\Socialite\Two\User` share. I also moved the declaration of the property to the abstract class because it's the same in both subclasses.

This doesn't break backwards compatibility so it should be pretty harmless to integrate in a minor version.

### DIFF
--- a/src/AbstractUser.php
+++ b/src/AbstractUser.php
@@ -43,6 +43,13 @@ abstract class AbstractUser implements ArrayAccess, User
     public $avatar;
 
     /**
+     * The user's access token.
+     *
+     * @var string
+     */
+    public $token;
+
+    /**
      * The user's raw attributes.
      *
      * @var array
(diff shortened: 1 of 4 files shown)


taylorotwell:
Hey Jose,

Is this fixing a bug?

josepostiga:
Hi Taylor!

As far as I know, it's not fixing a bug. I'm working on a project where I'm using Laravel Socialite to create user accounts, and as far as the project cares, it doesn't really matter which concrete type of `User` is used, as both of them will have the property, but it would be nice to not have to depend on a concrete subclass of `User` to not have my IDE screaming at me because I'm "accessing an undefined property".

😅

josepostiga:
Oh, right. Forgot the small detail of anyone that could have other implementations of this contract in their own projects. 

But I still think we could benefit from this. Would you be open to have a major version? Maybe taking the opportunity to also do some refactoring by adding types to methods? 😊 

If you see any value in that, I could try and help out in a fork and send it over for review? Maybe send over some notes over tweaks we could do besides this one, before anything. If the team likes it, we all win. If not, no harm no fall…

But I’d totally understand if that’s not something you guys would be open to invest on, at this time.

PR #658, opened 2023-08-31
Update LinkedInProvider.php for new LI OpenId

stephenstack:
See details on 

Sign in with LinkedIn is deprecated, here you have the new OpenId method #1054
https://github.com/SocialiteProviders/Providers/issues/1054

Linkedin changed its authentication method from Sign in with LinkedIn to Sign in with LinkedIn using OpenID Connect, since then the new app registered recently had this error.


stephenstack:
Lazy PR on my part, will update proper with tests and resubmit

PR #659, opened 2023-09-03
[5.x] Updated LinkedIn Provider to use the new Sign In with LinkedIn using OpenID Connect

joniand:
As of August 1, 2023 LinkedIn deprecated the Sign In with LinkedIn API endpoints in favour of the new [Sign In with LinkedIn using OpenID Connect](https://learn.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin-v2).

This PR makes the necessary changes to the LinkedInProvider.php to get it working again.


src/Two/LinkedInProvider.php
```diff
@@ -24,99 +24,64 @@ class LinkedInProvider extends AbstractProvider implements ProviderInterface
...
-            'id' => $user['id'],
+            'id' => $user['sub'],
             'nickname' => null,
-            'name' => $firstName.' '.$lastName,
-            'first_name' => $firstName,
-            'last_name' => $lastName,
-            'email' => Arr::get($user, 'emailAddress'),
-            'avatar' => Arr::get($avatar, 'identifiers.0.identifier'),
-            'avatar_original' => Arr::get($originalAvatar, 'identifiers.0.identifier'),
+            'name' => $user['name'].' '.$user['family_name'],
```
sajaddp, line 79: in this part:

`'name' => $user['name'].' '.$user['family_name'],`

Should it be `$user['name']` or `$user['given_name']?

joniand: Indeed, it should be $user['given_name']. Thanks

sajaddp:
@joniand I think a test is needed to confirm the PR.

joniand:
Thing is, there are no tests for any other providers. 

There was one for LinkedIn but from my understanding it was because there were 2 calls made, one for the details and one for the email which is not the case anymore.

PR #660, opened 2023-09-04
Updated LinkedIn Provider to use the new Sign In with LinkedIn using OpenID Connect

joniand:
As of August 1, 2023 LinkedIn deprecated the Sign In with LinkedIn API endpoints in favour of the new [Sign In with LinkedIn using OpenID Connect](https://learn.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin-v2).

This PR makes the necessary changes to the LinkedInProvider.php to get it working again.


nunomaduro:
@joniand If we were to release a version 6.x of socialite, what would be the upgrade guide for existing customers? Do they have to change anything? Or change anything at Linkedin level?

nunomaduro:
Going to rework the PR a bit on [feat/linkedin-openid-connect](https://github.com/laravel/socialite/tree/feat/linkedin-openid-connect).

joniand:
@nunomaduro there are no breaking changes to the response. 

In fact it is why I kept "avatar_original", as I didn't want to cause any issues with current implementations. If you are going to use it on a new release I'd be in favour of getting rid of it.

On the LinkedIn's side, I don't know if legacy implementations are migrated to the new "Sign In with LinkedIn using OpenID Connect" Product automatically or not and couldn't find anything on Microsoft's docs referencing it.

In my opinion mentioning that "Sign In with LinkedIn using OpenID Connect" must be activate should be enough.

Regarding the other pull request, it seems almost identical to mine with the difference that I've removed the obsolete test as well. That being said, I haven't tested it myself so can't say for sure if it works or not...

PR #667, opened 2023-10-05
Updating the api for getting emails as previous one not working for github provider

lmottasin:
When we are getting emails list with previous api for github.  When we are getting by user
`use Laravel\Socialite\Facades\Socialite; `
`$user = Socialite::driver('github')->userFromToken($token);`
So, I checked the api call manually with valid token still not getting the email. Then in github docs find another api endpoint to get the public emails. So just changing the urls `https://../user/emails to https://../user/public_emails` it fixed the problem.


taylorotwell:
Need more community feedback here.

PR #675, opened 2023-11-24
[5.x] Add refreshToken method

antoinelame:
This PR introduces a `refreshToken` method for OAuth 2.0 providers.

Typically, access tokens are updated in the database each time a user logs in. However, this approach may not be adequate for applications that perform actions on behalf of users, such as fetching data from social channels. There are scenarios where the user's session outlasts the token's validity, or where the application needs to use the token without the user's active login.

To address this, the proposed `refreshToken` method enables the application to refresh access tokens at any time using stored refresh tokens. 

```php
$newToken = Socialite::driver('gitlab')->refreshToken($user->gitlab_refresh_token);

$user->update([
    'gitlab_token' => $newToken->token,
    'gitlab_refresh_token' => $newToken->refreshToken,
    'gitlab_token_expires_at' => now()->addSeconds($newToken->expiresIn),
]);
```

The method returns an instance of the `Token` class, which includes the refreshed access token, refresh token, expiration duration, and scope.

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -341,6 +341,33 @@ abstract class AbstractProvider implements ProviderContract
     }
 
     /**
+     * Refresh a user access token with a refresh token.
+     *
+     * @param  string  $refreshToken
+     * @return Token
+     */
+    public function refreshToken($refreshToken)
+    {
+        $response = $this->getHttpClient()->post($this->getTokenUrl(), [
+            RequestOptions::HEADERS => ['Accept' => 'application/json'],
+            RequestOptions::FORM_PARAMS => [
+                'grant_type' => 'refresh_token',
+                'refresh_token' => $refreshToken,
+                'client_id' => $this->clientId,
+                'client_secret' => $this->clientSecret,
+            ]
+        ]);
+
+        $response = json_decode($response->getBody(), true);
+
+        return (new Token)
+            ->setToken(Arr::get($response, 'access_token'))
+            ->setRefreshToken(Arr::get($response, 'refresh_token'))
+            ->setExpiresIn(Arr::get($response, 'expires_in'))
+            ->setApprovedScopes(explode($this->scopeSeparator, Arr::get($response, 'scope', '')));
+    }
+
+    /**
      * Get the code from the request.
      *
      * @return string
--- a/src/Two/Token.php
+++ b/src/Two/Token.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace Laravel\Socialite\Two;
+
+class Token
+{
+    /**
+     * The user's access token.
+     *
+     * @var string
+     */
+    public $token;
+
+    /**
+     * The refresh token that can be exchanged for a new access token.
+     *
+     * @var string
+     */
+    public $refreshToken;
+
+    /**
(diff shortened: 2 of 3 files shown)


taylorotwell:
Does this code work for every first-party supported OAuth2 provider?

antoinelame:
Give me 5 days, I will test them all.

antoinelame:
## Bitbucket
✅ Works well.

## Slack
✅ Works well.

## Twitter
✅ Works well.

## GitLab
✅ Works well.

## Google
✅ Works well.

Note: to get a refresh token , you must add `['access_type' => 'offline']` to `getTokenFields`. So you'll have to tweak Socialite a little. But once you've got it, you can pass it to `refreshToken` method without any issue.

## LinkedIn
✅ Would work well.

I've not been able to create an application authorized to receive refresh tokens. It's reserved for a [limited number of partners](https://learn.microsoft.com/en-us/linkedin/shared/authentication/authorization-code-flow?context=linkedin%2Fconsumer%2Fcontext&view=li-lms-2022-07&tabs=HTTPS1#step-5-refresh-access-token). But it would work. The endpoint expects the [same request body parameters](https://learn.microsoft.com/en-us/linkedin/shared/authentication/programmatic-refresh-tokens#step-2-exchanging-a-refresh-token-for-a-new-access-token) as those sent by the proposed `refreshToken` method.

## Facebook

Facebook doesn’t provide refresh tokens. The `refreshToken` method is not relevant here.

## GitHub

For OAuth Apps, access tokens don’t expire. Refresh tokens [can only be enabled for GitHub Apps](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens#configuring-your-app-to-use-user-access-tokens-that-expire). The `refreshToken` method is not relevant here.

NileshS-RSW:
I facing this problem
"laravel/socialite": "^5.5",
"socialiteproviders/apple": "^5.5",

```
Declaration of SocialiteProviders\\Apple\\Provider::refreshToken(string $refreshToken): Psr\\Http\\Message\\ResponseInterface must be compatible with Laravel\\Socialite\\Two\\AbstractProvider::refreshToken($refreshToken)',
  'file' => '/var/www/html/coinpanda-backend/vendor/socialiteproviders/apple/Provider.phpv
```
-> Can you help me to solve this problem

driesvints:
@NileshS-RSW report this issue to the socialite providers repo.

VirusEcks:
apple is broken too because of this update

pawell67:
Okta is broken as well

cosmastech:
Seems like this broke a lot of things. Any chance this might get reverted? @taylorotwell

antoinelame:
Hey @cosmastech,

It's unfortunate that the update affected some third-party providers.

However, I believe it's important to be able to  evolve Socialite without being restricted by third-party providers. You can submit a Pull Request to [SocialiteProviders/Providers](https://github.com/SocialiteProviders/Providers). The required change is quite straightforward: the `refreshToken` method doesn't need to be declared on the third-party provider, as it's now already included in the `AbstractProvider`.

In the meantime, while waiting for your PR, or someone else's, to be accepted, you can do this:

```
composer require laravel/socialite:5.10.0
```

PR #680, opened 2023-12-25
Added expiresAt Carbon attribute to OAuth2 token class

dmyers:
This is a helper attribute which I've added in my own usage of Socialite when working with `expires_at` which are plain integers.

Having a `Carbon` object readily available is useful when storing tokens in a database and working with refreshing tokens in commands.


taylorotwell:

PR #688, opened 2024-02-09
fix(google provider): check if response body of GoogleProvider's getU…

alibori:
…serByToken method is a GuzzleHttp\Psr7\Stream instance

Fix `TypeError` error caused on `GoogleProvider`'s `getUserByToken` method when trying to `json_decode` the response body.


alibori:
Ok! I'll try!

PR #689, opened 2024-02-10
Correct issue with GoogleProvider getUserByToken() method

alibori:
Calling `getUserByToken` from `GoogleProvider` crashes with:

```bash
TypeError error json_decode(): Argument #1 ($json) must be of type string, GuzzleHttp\Psr7\Stream given.
```

To handle this situation, added this code in `getUserByToken` method:

```php
  if ($response->getBody() instanceof Stream) {
      return json_decode($response->getBody()->getContents(), true);
  }
```

Finally, test added to check method works as expected.

### DIFF
--- a/src/Two/GoogleProvider.php
+++ b/src/Two/GoogleProvider.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Socialite\Two;
 
+use GuzzleHttp\Psr7\Stream;
 use GuzzleHttp\RequestOptions;
 use Illuminate\Support\Arr;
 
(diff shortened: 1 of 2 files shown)


driesvints (approved):

iBotPeaches (commented):
I don't understand how this is happening - wouldn't this affect all providers? What makes the Google one special?

Just saying this as I just pulled dev-master to get the last merged PR to test my fixes and I've never once replicated this error. Can't figure out what I'm missing.

taylorotwell:
Needs reproduction example.

driesvints:
@alibori no worries we merged this one 👍

PR #698, opened 2024-04-12
Support OIDC token in Facebook provider (Facebook Limited Login)

gdebrauwer:
Facebook recently made changes to its Facebook Login on iOS: https://developers.facebook.com/blog/post/2024/03/28/changes-made-to-fb-login-sdk/

When a user does not allow app tracking, the login uses a new 'limited login' flow that returns an OIDC token. This token can not be used to access Facebook's Graph API. That means the current Facebook provider does not work in this 'limited login' flow. Apple requires you to use Facebook's latest SDK version.  That SDK version uses this new 'limited login' flow. This makes the current Facebook socialite provider unusable when validating a Facebook login via iOS SDK.

This PR fixes that by adding support for the OIDC token in a way that does not break or impact the regular Facebook login.

I implemented the validation of the OIDC token based on Facebook's documentation: https://developers.facebook.com/docs/facebook-login/limited-login/token/validating

Example of a user returned from the `userFromToken()` method when an OIDC token was provided:
Screenshot 2024-04-12 at 13 49 59

I did not add tests as I don't really see a way how I can write a test for this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,11 +17,13 @@
     "require": {
         "php": "^7.2|^8.0",
         "ext-json": "*",
+        "firebase/php-jwt": "^6.10",
         "guzzlehttp/guzzle": "^6.0|^7.0",
         "illuminate/contracts": "^6.0|^7.0|^8.0|^9.0|^10.0|^11.0",
         "illuminate/http": "^6.0|^7.0|^8.0|^9.0|^10.0|^11.0",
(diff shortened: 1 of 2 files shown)


driesvints:
@gdebrauwer tests break here

taylorotwell:
Code also needs docblocks.

driesvints:
Thanks @gdebrauwer. Do we need a PR to the docs as well for this?

gdebrauwer:
@driesvints I created a docs PR

PR #709, opened 2024-06-28
Check if some fields exist using limited login

CosmicodeStudio:


CosmicodeStudio:
Can you please merge this? It's crashing on all users when using Facebook Limited Login in iOS

PR #710, opened 2024-07-05
Additional Getter Methods For Retrieving User Information

jeramyhing:
## Description

This PR updates the `Laravel\Socialite\AbstractUser` and `Laravel\Socialite\Two\User` classes to add additional getter methods. It also enhances the existing tests to verify that the new getter methods return the correct property values.

### `AbstractUser` Class Updates

**New Getter Method:**

- `getAttributes`: Get the user's attributes.

### `Two\User` Class Updates:

**New Getter Methods:**

- `getToken`: Retrieves the user's access token.
- `getRefreshToken`: Retrieves the user's refresh token.
- `getExpiresIn`: Retrieves the number of seconds the access token is valid for.
- `getApprovedScopes`: Retrieves the scopes approved by the user during authentication.

### Tests:

- Updated the `OAuthOneTest` and `OAuthTwoTest` class to use the new getter methods.
- Added assertions to verify that the getter methods return the correct property values.

## Benefits to End Users

These updates provide more convenient methods for retrieving user information, making it easier to work with user data. The changes are backward-compatible and do not break any existing features. Note that while the `Two\User` getters enhance data retrieval, the properties still require type-hinting for full static analysis compatibility.

### DIFF
--- a/src/AbstractUser.php
+++ b/src/AbstractUser.php
@@ -130,6 +130,16 @@ abstract class AbstractUser implements ArrayAccess, User
     }
 
     /**
+     * Get the user's attributes.
+     *
+     * @return array
+     */
+    public function getAttributes()
+    {
+        return $this->attributes;
+    }
+
+    /**
      * Map the given array onto the user's properties.
      *
      * @param  array  $attributes
--- a/src/Two/User.php
+++ b/src/Two/User.php
@@ -35,6 +35,16 @@ class User extends AbstractUser
     public $approvedScopes;
 
     /**
+     * Get the access token for the user.
+     *
+     * @return string
+     */
+    public function getToken()
+    {
+        return $this->token;
+    }
+
+    /**
      * Set the token on the user.
      *
      * @param  string  $token
@@ -45,6 +55,16 @@ class User extends AbstractUser
         $this->token = $token;
 
         return $this;
+    }
+
+    /**
+     * Get the refresh token for the user.
+     *
+     * @return string
+     */
+    public function getRefreshToken()
+    {
+        return $this->refreshToken;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


src/Contracts/User.php
```diff
@@ -38,4 +38,18 @@ public function getEmail();
...
+     * @return array
+     */
+    public function getRaw();
+
+    /**
+     * Get the user's attributes.
+     *
+     * @return array
+     */
+    public function getAttributes();
```
driesvints, line 54: We can't add methods to interfaces on patch releases. This will break all apps implementing custom providers.

driesvints:
Please revert the changes to the interface and mark for review again. Thanks.

src/Contracts/User.php
jeramyhing: Of course, I will revert the interface. Is there a major release branch that these additions can go on?

driesvints: Maybe we can add it as `@method` annotations for now?

taylorotwell:
I don't think we need all this for public properties?

src/Contracts/User.php
jeramyhing: Yeah, I considered that. Seems to do the trick on my local branch. The whole PR came about as a step toward avoiding direct type-hinting with inline PHPDocs, due to PHPStan not recognizing the public properties:

```php
/** @var \Laravel\Socialite\Two\User $user */
$user = Socialite::driver('facebook')->user();

$attributes = $user->attributes;
```

With the proposed changes it would simply be:

```php
$user = Socialite::driver('facebook')->user();

$attributes = $user->getAttributes();
```

This proposed change leverages new getters like the proposed `getAttributes()` method, which is a more explicit and likely better practice for accessing the user’s attributes. As a method, it could also be added to the class PHPDoc as you had suggested. This would make for better static analysis support, since properties aren't meant for interfaces. The change was aimed at making PHPStan recognize the attributes correctly without relying on the type hint.

However, I can see that @taylorotwell doesn't think that is needed. So, this is all a moot issue now. I'll just revert my development back to type-hinting the `$user` var to make PHPStan happy.

Thanks guys.

PR #747, opened 2025-06-21
Improve error handling for invalid access token responses

TobMoeller:
When an OAuth provider's token endpoint returns an invalid response (non-JSON or missing access_token), 
the current code fails with a generic error. This can be confusing when debugging issues, 
especially in cases where firewalls or proxy servers intercept the request and return HTML instead 
(e.g., Cloudflare access denied pages).

This PR adds an `InvalidTokenResponseException` that is thrown, if the OAuth provider's response does not contain the required `access_token`.

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -239,7 +239,7 @@ abstract class AbstractProvider implements ProviderContract
 
         $response = $this->getAccessTokenResponse($this->getCode());
 
-        if (! Arr::has($response, 'access_token')) {
+        if (! is_array($response) || ! Arr::has($response, 'access_token')) {
             throw new InvalidTokenResponseException;
         }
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
I wonder if the response could be included in the exception? Something that might let developers inspect what is going on.

TobMoeller:
Thanks for the suggestion!

I looked into including the response in the exception, but unfortunately, by the time the exception is thrown, the response has already been processed through json_decode(). This means that any response not containing valid JSON (e.g., an HTML error page or plain text) ends up as null. As a result, we don't have the original response available when raising the exception.

Altering the behavior of the `getAccessTokenResponse` method or raising the exception earlier would likely result in a breaking change.

If you have other thoughts or an alternative approach, please let me know—I’d be happy to explore it further!

Cellard:
I think that this PR should cover not only `access_token` error response, as it discribed in https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

but should also handle errors in redirect callback, as described in https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

The nature of these errors in both scenarios is very similar.

PR #764, opened 2026-02-26
Rename Boost skill directory to socialite-development

pushpak1300:
Renames the Boost skill directory from `socialite` to `socialite-development`


JoshSalway:
What benefits does this have?

pushpak1300:
@JoshSalway consistency.

PR #767, opened 2026-03-18
Fix PKCE session usage in stateless mode

### LINKED ISSUE #754: Session issue when using Laravel Socialite X (Twitter) in stateless mode
### Socialite Version

5.23.1

### Laravel Version

12.35.1

### PHP Version

8.4.14

### Database Driver & Version

_No response_

## 1. Issue Occurred
- Reason: The problem occurred because a stateless process, which shouldn't save state, was using sessions.
- Location of the Issue: vendor/laravel/socialite/src/Two/AbstractProvider.php@redirect()
- Lines: 169 ~ 171
- Code:
```php
public function redirect()
{
    $state = null;

    if ($this->usesState()) {
        $this->request->session()->put('state', $state = $this->getState());
    }

    if ($this->usesPKCE()) {
        $this->request->session()->put('code_verifier', $this->getCodeVerifier());
    }

    return new RedirectResponse($this->getAuthUrl($state));
}
```

## 2. Cause
- Location: vendor/laravel/socialite/src/Two/TwitterProvider.php
- Line: 22
- Code:
```php
protected $usesPKCE = true;
```

## 3. Solution
1. Provider Creation
```php
<?php

namespace App\Services;

use Laravel\Socialite\Two\XProvider as BaseXProvider;

class CustomXProvider extends BaseXProvider
{
    /**
     * Forces the simplest **Stateless** mode by disabling PKCE usage,
     * which avoids the use of session-related logic.
     *
     * @var bool
     */
    protected $usesPKCE = false;
}
```
2. Provider Registration
```php
$socialite = $this->app->make(SocialiteFactory::class);
// Extend the 'x' (or 'twitter') driver with the custom driver.
$socialite->extend(
    'x',
    function ($app) use ($socialite) {
        // Inject the 'x' configuration from config/services.php into the custom provider.
        $config = $app['config']['services.x'];

        // Create and return an instance of CustomXProvider.
        return $socialite->buildProvider(CustomXProvider::class, $config);
    }
);
```
3. Resolution

### Steps To Reproduce

It occurs when installing according to the manual and using X's OAuth 2.0 in stateless mode.

JoshSalway:
## Summary
- When a provider uses PKCE (e.g. Twitter/X) and `stateless()` is enabled, PKCE operations attempted to access the session, causing exceptions since no session is available
- Guards all three PKCE session operations (`redirect()`, `getCodeFields()`, `getTokenFields()`) with the existing `usesState()` check, so PKCE is only used when session state is available
- This is consistent with how `state` is already handled -- only stored/validated when `usesState()` returns true

Fixes #754

## Test plan
- [x] Added test: redirect with PKCE provider in stateless mode does not use session or throw exceptions
- [x] Added test: token request excludes `code_verifier` in stateless mode
- [x] Added test: auth URL excludes `code_challenge` parameters in stateless mode
- [x] All 17 existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


taylorotwell:
These PRs are too AI generated and it's making it hard for me to sus out what is really an issue and thought through and considered and what is not.

PR #768, opened 2026-03-18
Make redirect config key optional when using redirectUrl()

### LINKED ISSUE #740: Using redirectUrl still needs redirect key in config
### Socialite Version

5.18

### Laravel Version

12

### PHP Version

8.4

### Database Driver & Version

_No response_

### Description

When I set `redirectUrl()` still need set `redirect` key inside config/services.php.
Why?

Now I set empty string.
It will be good to `redirect` not should be required when using `recirectUrl()` method.

### Steps To Reproduce

Use `redirectUrl()` and remove `redirect` key from config/services.php

JoshSalway:
## Summary
- Removes `redirect` from the required config keys in `buildProvider()`, allowing users to set the redirect URL programmatically via `redirectUrl()` without needing an empty `redirect` key in `config/services.php`
- Handles missing `redirect` key gracefully in `formatRedirectUrl()` by defaulting to null

Fixes #740

## Test plan
- [x] Updated test: driver can be instantiated without `redirect` config key
- [x] Updated test: exception message for completely missing config reflects new required keys
- [x] All non-JWT tests pass (JWT failures are pre-existing due to missing firebase/php-jwt dependency)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -239,7 +239,7 @@ class SocialiteManager extends Manager implements Contracts\Factory
      */
     public function buildProvider($provider, $config)
     {
-        $requiredKeys = ['client_id', 'client_secret', 'redirect'];
+        $requiredKeys = ['client_id', 'client_secret'];
 
         $missingKeys = array_diff($requiredKeys, array_keys($config ?? []));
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Original issue isn't clear tbh.

PR #771, opened 2026-04-24
Fix usage of access_token in Bitbucket integration

fetzi:
The Bitbucket integration is currently failing because they deprecated sending access tokens via GET or POST parameters. Instead the access token needs to be sent as a Bearer Token.

See https://developer.atlassian.com/cloud/bitbucket/changelog/#OAuth-access-tokens-can-no-longer-be-provided-via-query-parameters-or-POST-body for more details.

This PR replaces the query parameter `access_token` with the Bearer token for user and e-mail retrieval.


fetzi:
@taylorotwell Could you please publish this fix? The integration with Bitbucket is currently not working.

PR #776, opened 2026-06-09
[5.x] Add ability to easily fake Socialite users

stevebauman:
**Description**:

This PR adds a convenient way to create fake Socialite user instances when testing Socialite callbacks.

**Usage**:

Instead of manually mapping the expected user attributes, developers may now create fake OAuth users directly from the appropriate Socialite user classes themselves and save on having to define them all and instead override what they need overridden. For example:

Before:

```php
use Laravel\Socialite\Two\User;
use Laravel\Socialite\Socialite;

$attributes = [
    'id' => '123456789',
    'nickname' => 'testuser',
    'name' => 'Test User',
    'email' => 'test@example.com',
    'avatar' => 'https://example.com/avatar.jpg',
];

Socialite::fake('github', (new User)->setRaw($attributes)->map($attributes)
    ->setToken('fake-token')
    ->setRefreshToken('fake-refresh-token')
    ->setExpiresIn(3600)
    ->setApprovedScopes([]));
```

After:

```php
use Laravel\Socialite\Two\User;
use Laravel\Socialite\Socialite;

Socialite::fake('github', User::fake([
    'email' => 'test@example.com',
]));
```

OAuth 1 users may also be faked:

```php
use Laravel\Socialite\One\User;
use Laravel\Socialite\Socialite;

Socialite::fake('twitter', User::fake([
    'nickname' => 'testuser',
]));
```

### DIFF
--- a/src/Testing/SocialiteFake.php
+++ b/src/Testing/SocialiteFake.php
@@ -45,7 +45,7 @@ class SocialiteFake implements Factory
      * Register a fake user for the given driver.
      *
      * @param  string  $driver
-     * @param  \Laravel\Socialite\Contracts\User|\Closure|array|null  $user
+     * @param  \Laravel\Socialite\Contracts\User|\Closure|null  $user
      * @return $this
      */
     public function fake($driver, $user = null)
--- a/src/One/User.php
+++ b/src/One/User.php
@@ -24,7 +24,7 @@ class User extends AbstractUser
      * Create a fake OAuth 1 user instance.
      *
      * @param  array  $attributes
-     * @return static
+     * @return self
      */
     public static function fake(array $attributes = [])
     {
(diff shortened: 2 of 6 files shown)


src/Testing/SocialiteFake.php
```diff
@@ -45,7 +45,7 @@ public function driver($driver = null)
      * Register a fake user for the given driver.
      *
      * @param  string  $driver
-     * @param  \Laravel\Socialite\Contracts\User|\Closure|array|null  $user
+     * @param  \Laravel\Socialite\Contracts\User|\Closure|null  $user
```
stevebauman, line 48: I've updated these annotations because `array`'s are not actually supported.

We can't safely map a `User` socialite instance when provided an array since we don't know what version of OAuth the provider is to create the correct one (`Two\User` | `One\User`).

PR #787, opened 2026-07-08
Make the `redirect` config key optional

### LINKED ISSUE #740: Using redirectUrl still needs redirect key in config

nikolauspschuetz:
Closes #740.

### Problem

The redirect URL can be set at runtime with `->redirectUrl($url)`, but `SocialiteManager::buildProvider()` still lists `redirect` as a **required** config key. So even when you only ever set the redirect programmatically, you have to add a placeholder `redirect` (e.g. an empty string) to `config/services.php`, or you get a `DriverMissingConfigurationException`.

### Change

- Drop `redirect` from `$requiredKeys` in `buildProvider()` (only `client_id` and `client_secret` remain required).
- Make `formatRedirectUrl()` tolerate a missing `redirect` key (`$config['redirect'] ?? null`).

You can now do:

```php
// config/services.php — no redirect key needed
'github' => [
    'client_id'     => env('GITHUB_CLIENT_ID'),
    'client_secret' => env('GITHUB_CLIENT_SECRET'),
],
```
```php
Socialite::driver('github')->redirectUrl($url)->redirect();
```

### Note on behavior change

This relaxes a previously-required (and tested) key, so I updated `SocialiteManagerTest`:

- `test_it_throws_exception_when_redirect_is_missing` → `test_it_can_instantiate_the_driver_without_redirect_configuration` (now asserts the provider builds).
- `test_it_throws_exception_when_configuration_is_completely_missing` — expected missing-keys message updated to `[client_id, client_secret]`.

Full suite passes (48 tests). Happy to adjust if you'd prefer to keep `redirect` required — this follows the `help wanted` label and the request in #740.

### DIFF
--- a/src/SocialiteManager.php
+++ b/src/SocialiteManager.php
@@ -239,7 +239,7 @@ class SocialiteManager extends Manager implements Contracts\Factory
      */
     public function buildProvider($provider, $config)
     {
-        $requiredKeys = ['client_id', 'client_secret', 'redirect'];
+        $requiredKeys = ['client_id', 'client_secret'];
 
         $missingKeys = array_diff($requiredKeys, array_keys($config ?? []));
 
@@ -277,7 +277,7 @@ class SocialiteManager extends Manager implements Contracts\Factory
      */
     protected function formatRedirectUrl(array $config)
     {
-        $redirect = value($config['redirect']);
+        $redirect = value($config['redirect'] ?? null);
 
         return Str::startsWith($redirect ?? '', '/')
                     ? $this->container->make('url')->to($redirect)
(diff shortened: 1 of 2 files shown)


taylorotwell:

If applicable, please consider releasing your code as a package so that the community can still take advantage of your contributions!

PR #789, opened 2026-08-13
fix: Validate Facebook OIDC token nonce

meihao550:
- I added nonce validation because token was not properly checked
- I sent a message for Taylor a month ago but I couldn't get any reply. so here it is

### DIFF
--- a/src/Two/AbstractProvider.php
+++ b/src/Two/AbstractProvider.php
@@ -262,7 +262,7 @@ abstract class AbstractProvider implements ProviderContract
     }
 
     /**
-     * Get a Social User instance from a known access token.
+     * Get a Socialite user instance from a known access token.
      *
      * @param  string  $token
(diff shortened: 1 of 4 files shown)


meihao550:
Hi Taylor,

A while back I emailed this address about the Facebook provider not
validating the nonce claim on Limited Login OIDC tokens. I didn't hear
back, so I opened it as a public PR (#789), which you merged on August 13.

Because the omission allowed a valid id_token to be replayed into an
unrelated authentication session, I'm planning to request a CVE ID for it
through VulnCheck.

Please let me know if you have any objection, or if Laravel would rather
handle the assignment directly — I'm happy to hold off or defer to you.

Best regards,
[MY NAME]
GitHub: meihao550


2026年8月14日(金) 8:01 Taylor Otwell ***@***.***>:

PR #794, opened 2026-09-19
Use 1.12 for league/oauth1-client

Universal-Omega:
1.12 was just released with Guzzle 8 support


composer.json
```diff
@@ -25,7 +25,7 @@
         "illuminate/contracts": "^6.0|^7.0|^8.0|^9.0|^10.0|^11.0|^12.0|^13.0",
         "illuminate/http": "^6.0|^7.0|^8.0|^9.0|^10.0|^11.0|^12.0|^13.0",
         "illuminate/support": "^6.0|^7.0|^8.0|^9.0|^10.0|^11.0|^12.0|^13.0",
-        "league/oauth1-client": "^1.11",
```
crynobone, line 28: `^1.11` already allows `1.12.0` to be installed. No changes is required.

Universal-Omega: I'm aware of that, but since this was the latest version before I assumed that was the intent that it remain the latest version so that's why I did this. If that isn't the case this can be closed. I should have clarified this in commit message also.

Universal-Omega: Anyway thanks for the clarification. Glad to finally get it to work with Guzzle 8 anyway.

